PR #1128, opened 2014-03-15
Add cpancover links where appropriate

oalders:
I think the JSON caching is "good enough" for our purposes right now.  I've never used this particular CHI driver in production before, though, so wanted to know if anyone had any concerns with using it.  Thought it would be faster than using something on disk and the memory required shouldn't be much.  The JSON file is around 43 kb right now, but it should grow as more dists get added to the nightly run.

This is all pretty new (the JSON summary was just implemented yesterday), but it seems this will be updated every 24 hours or so going forward.  Also, old reports may not get cached, so we need to be up to date on what is available.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -65,6 +65,7 @@ requires 'Template::Plugin::Number::Format';
 requires 'Template::Plugin::Page';
 requires 'Try::Tiny', '0.09';
 requires 'URI';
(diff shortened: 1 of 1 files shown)


mohawk2:
@oalders Can this branch be zapped?

PR #1438, opened 2015-01-09
elasticsearch 1.x field values shim

mateu:
elasticsearch field values are always returned as arrays starting with version 1.0.  This code handles the extraction of the first value in the array for 1.x and leaves pre 1.0 results alone

### DIFF
--- a/lib/MetaCPAN/Web/Role/Elasticsearch/Adapter.pm
+++ b/lib/MetaCPAN/Web/Role/Elasticsearch/Adapter.pm
@@ -1,4 +1,4 @@
-package MetaCPAN::Web::Role::Adapter;
+package MetaCPAN::Web::Role::Elasticsearch::Adapter;
 
 use Moose::Role;
 
--- a/lib/MetaCPAN/Web/Role/Elasticsearch/Adapter.pm
+++ b/lib/MetaCPAN/Web/Role/Elasticsearch/Adapter.pm
@@ -4,19 +4,93 @@ use Moose::Role;
 
 =head1 METHODS
 
-=head2 extract_first_element
+=head2 single_valued_arrayref_to_scalar
 
 Elasticsearch 1.x changed the data structure returned when fields are used.
 For example before one could get a ArrayRef[HashRef[Str]] where now
 that will come in the form of ArrayRef[HashRef[ArrayRef[Str]]]
 
-This function will select the first element given an array of hash of array.
+This function reverses that behavior
+By default it will do that for all fields that are a single valued array,
+but one may pass in a list of fields to restrict this behavior only to the
+fields given.
+
+So this:
+
+    $self->single_valued_arrayref_to_scalar(
+    [
+      {
+        name     => ['WhizzBang'],
+        provides => ['Food', 'Bar'],
+      },
+       ...
+   ]);
+
+yields:
+
+    [
+      {
+        name     => 'WhizzBang',
+        provides => ['Food', 'Bar'],
+      },
+      ...
+    ]
+
+and this estrictive example):
+
+    $self->single_valued_arrayref_to_scalar(
+    [
+      {
+        name     => ['WhizzBang'],
+        provides => ['Food'],
+      },
+       ...
+   ], ['name']);
+
+yields:
+
+    [
+      {
+        name     => 'WhizzBang',
+        provides => ['Food'],
+      },
+      ...
+    ]
+
+=cut
+
+sub single_valued_arrayref_to_scalar {
+    my ( $self, $array, $fields ) = @_;
+    my $has_fields = defined $fields ? 1 : 0;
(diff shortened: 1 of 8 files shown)


lib/MetaCPAN/Web/Role/Elasticsearch/Adapter.pm
```diff
@@ -0,0 +1,50 @@
...
+=cut
+
+sub extract_first_element {
+    my ( $self, $array ) = @_;
+    foreach my $hash ( @{$array} ) {
+        foreach my $field ( %{$hash} ) {
+            my $value = $hash->{$field};
+
+            # We only extract the first element when have an ArrayRef
+            next if not( ref($value) and ( ref($value) eq 'ARRAY' ) );
```
rwstauner: Should this be testing the length of the array?
You don't want to discard other elements if there are multiple.
Or is this only intended to be used for certain queries?
I see that the current places where it's used there shouldn't be any multi-element arrays.

mateu: It's a good point you bring up.  Yes it's intended only for single element
arrays where the field value was just a scalar before.   However, I'm open
to adding a count check and die if the value encountered is a multi-element
array?

2015-01-09 10:31 GMT-07:00 Randy Stauner <notifications@github.com>:

rwstauner: I don't think a runtime error is the answer...
This is intended for specific queries (and explicitly not useful for specific other queries),
so the issue as I see it is the likelihood for the developer to make a mistake.

If the developer is going to accidentally use this when they shouldn't, then it seems equally likely to me that they might test it on a value that only has one element, but if there's a value in production that has multiple, we'll be serving 500's.

Which is better?  A mysterious bug that shows only one of multiple values, or serving errors to the user?

Perhaps we should consider making the code forward compatible instead of backward compatible?

ranguard: For anyone not aware, we intend to have a totally separate domain (api.v1.metacpan.org or something) for the 1.x ES - this is because we will be putting it through Fastly AND dropping support for GET+Body.

We will need to maintain the old API for a while (timeframe to be decided) so if we can maintain this as 1 code base that would be great - but at worst case we can create a branch before merging breaking changes and fix the existing api.mc.org to that branch which we consider frozen.

oalders: I'd really like to avoid 2 branches as much as possible.  If this doesn't get completed in the OPfW term we're looking at another branch which could live as long as the bootstrap branch.  And trying to merge that made me sad.

What I like about this is that we could merge it into master _now_ and not have it sit in a long lived branch.

So, if an adapter is safe for us to use, I'd be more than happy to merge it into master and delete it after the upgrade.  Probably what we really would need to do in the meantime is write a few QA scripts so that we can compare query results on the 0.x and 1.x branches.

rwstauner: > So, if an adapter is safe for us to use, I'd be more than happy to merge
it into master and delete it after the upgrade.

Precisely why I suggested forward compatible rather than backward
compatible.  If we upgrade scalars to arrays (and fix the other code to
expect arrays) the code will be more future proof and the adapter can be
removed after the upgrade. If we instead downgrade arrays to scalars then
the adapter lives forever.

oalders:
@mateu what do you think about @rwstauner's last comment.  Sounds reasonable to me.

mateu:
I think it makes good sense to think forward compatible, but in the end I'm
not sure how much it matters because ultimately we'll have to extract the
single values from the arrays to work with the data in the controller or
view.  i.e. the shim will live on in some form or another regardless of the
route we choose.  One disadvantage is that it requires more up-front work,
because the code that currently expects a scalar will have to get it from
the (upgraded) array.   One advantage is that we can blindly update all
scalars to arrays (when using fields) where the current version of code
would actually need an explicit list of fields to operate on so we're not
turning a single valued array into scalar when we shouldn't.  In the end,
I'm not strongly attached to either approach.




2015-01-15 6:45 GMT-07:00 Olaf Alders <notifications@github.com>:

ranguard:
Where is this pull at... are we ready to merge :) ?

oalders:
@mateu I think we can merge this.  Could you rebase out the merge commit first?

mateu:
I'll make this proper this weekend.

2015-02-10 21:01 GMT-07:00 Olaf Alders <notifications@github.com>:

mateu:
I believe she's ready for merger

2015-02-11 11:15 GMT-07:00 Mateu X. Hunter <hunter@missoula.org>:

oalders:
Thank you!

PR #1448, opened 2015-01-23
Use https where possible

mbeijen:
I noticed some pages on metacpan use http instead of https where there are https links available.

This PR changes the links to api.metacpan.org, explorer.metacpan.org and perl.org to use https.


oalders:
:+1: Thanks very much!

PR #1452, opened 2015-02-06
[WIP]don't list modules from other releases

andreeap:
This is a try at https://github.com/CPAN-API/cpan-api/issues/125. I solved the problem of showing the modules from other releases, but in the search page, it's still wrong and I am not sure where that comes from. Maybe that is still from the api?

### DIFF
--- a/root/release.html
+++ b/root/release.html
@@ -43,7 +43,15 @@ provides      = [];
 modules       = [];
 FOREACH file IN files;
   IF file.documentation && file.module;
-    modules.push(file);
+    found = 0;
+    FOREACH module IN file.module;
+      IF module.name != file.documentation;
+        found = found + 1;
+        IF found == 1;
+          modules.push(file);
+        END;
+      END;          
+    END;
     FOREACH module IN file.module;
       IF module.name != file.documentation && module.indexed && module.authorized;
         # There must be a better way...


root/release.html
```diff
@@ -43,7 +43,15 @@
...
 FOREACH file IN files;
   IF file.documentation && file.module;
-    modules.push(file);
+    found = 0;
+    FOREACH module IN file.module;
+      IF module.name != file.documentation;
+        found = found + 1;
+        IF found == 1;
+          modules.push(file);
+        END;
```
oalders: You could shorten this to:

```
IF module.name != file.documentation;
    modules.push(file);
END;
```

oalders:
Could you summarize how this fixes the problem?

andreeap:
The way it was before it added modules from another release that were listed as a documentation, but by checking just the ```IF module.name != file.documentation;``` and then add the file it will push the same module 3 times and this is why I did that little thing with found. Of course, maybe it's something that I don't understand and it could be easily solved in some other way.

oalders:
@rwstauner thoughts?

oalders:
@andreeap is this still a WIP?

oalders:
@andreeap ping?

rwstauner:
It took me a while to figure out what this actually doing.
This is not the right way to fix this.
There's a bug in the template logic (before this change) so that this new code just isn't being run,
which happens to make the undesired modules disappear.

I'm considering several other (much simpler changes) that will change the way things are displayed and likely spark more discussion about how this should really be handled.

rwstauner:
Fixing these two (similar) template bugs made a significant difference in how those modules are displayed for JSON and JSON-PP:
cd0b4d69f421722dc2ac2d03e29743dd88db1409
77ba2e6c66461b634703d32d96b88249e13e3f5f

rwstauner:
Thanks for your effort on this, it did help to identify and fix the overall issue!

Issue #1430, opened 2014-12-17
Negative number of non-PAUSE ++ers


choroba:
See also https://github.com/CPAN-API/metacpan-web/issues/1305
This time at [JSON::Parse](http://metacpan.org/pod/JSON::Parse)
![1](https://cloud.githubusercontent.com/assets/886196/5468522/751d7cf0-85d0-11e4-8bd0-eee72dc2d073.png)

oalders:
Math is hard.

![screenshot 2015-01-17 07 11 48](https://cloud.githubusercontent.com/assets/96205/5788780/35b77916-9e18-11e4-8bc6-61ca1b0322ed.png)

This highlights the fact that we need to be using the same logic to calculate the ++ in the badge.

Csson:
Every distribution I have seen that on has been ++ by both sharyanto and perlancar. Maybe that helps tracking it down.

oalders:
Looks like this was fixed via #1455.  Thanks @oiami :)

PR #1455, opened 2015-02-11
Set 0 to non-PAUSE user

oiami:
A small fix for #1430 , I'm not sure if it's the right fix though. Since we get the number of non-PAUSE user from total_plussers - PAUSE_user = non-PAUSE . I'm not sure why number of users with PAUSE id who ++ed the module sometimes it's more than number of total plussers. I guess that because those PAUSE users don't register for a user id with MetaCPAN?   So, I just assume that if the number of PAUSE users more than total plusser then just assume that nobody is non-PAUSE user can set 0 number. Please just close this PR if I'm wrong.

### DIFF
--- a/lib/MetaCPAN/Web/Model/API/Favorite.pm
+++ b/lib/MetaCPAN/Web/Model/API/Favorite.pm
@@ -137,6 +137,10 @@ sub find_plussers {
     # find total non pauseid users who have ++ed the dist.
     my $total_nonauthors = ( $total_plussers - $total_authors );
 
+    # number of pauseid users can be more than total plussers
+    # then set 0 to non pauseid users
+    $total_nonauthors = 0 if $total_nonauthors < 0;
+
     return (
         {
             plusser_authors => \@plusser_details,


rwstauner:
@oiami do you have a python background?
That ternary looks like it's backwards: "value if condition else other_value" :-)

If this was the solution I'd probably leave the original assignment and then just add a line like:
`$total_nonauthors = 0 if $total_nonauthors < 0;`
That way the calculation is only done once and you save some typing.

If the math is coming out wrong, there's probably a bug somewhere else, but I haven't taken the time to trace it.
This might be better than leaving it to display as obviously wrong.

oiami:
@rwstauner I don't really have python back ground but I think the variable names are too long so I just tried to use shorthand but it could be wrong. Sorry about that. :(  Do you want me change to logic or want to find out about the math things first ?

lib/MetaCPAN/Web/Model/API/Favorite.pm
```diff
@@ -135,7 +135,10 @@ sub find_plussers {
     my $total_authors = @plusser_details;
 
     # find total non pauseid users who have ++ed the dist.
-    my $total_nonauthors = ( $total_plussers - $total_authors );
+    my $total_nonauthors
+        = ( $total_plussers - $total_authors )
```
rwstauner: If this amount is a true value...

lib/MetaCPAN/Web/Model/API/Favorite.pm
```diff
@@ -135,7 +135,10 @@ sub find_plussers {
...
+        ? ( $total_plussers - $total_authors ) > 0
```
rwstauner: save a boolean (whether or not this calculation is greater than zero) to the variable.

I think that's not what you meant.

rwstauner:
I think fixing it to not display a confusing value (negative) is sufficient for now.

In the interest of maximum clarity I would suggest using the two statements, with a comment in between.  Not only is it more DRY in that the calculation is not being repeated, but it clearly demonstrates the sentiment that "this works around bugs by fixing an obviously incorrect display issue".
I would personally prefer something like this:

```
my $total_nonauthors = ( $total_plussers - $total_authors );
# Don't show a negative number if the count isn't accurate.
$total_nonauthors = 0 if $total_nonauthors < 0;
```

but I'm open to discussion.

oiami:
Ok, fixed, no objection for your solution :+1: . I'm completely misunderstood about the shorthand. Sorry again :worried:  , I didn't do much coding these days so I was a bit confuse I should make more contribution to bring back my programming skill before it's getting much worse :sweat:

rwstauner:
No worries.  Glad to see you around here anytime!

Only one thing: What about the comment?  :-)
I think a comment to concisely describe the discussion we've had here would be valuable (like I one I suggested above).
It's always good to leave a comment about "why" something is being done (or what else might be worth considering).

Thanks for helping out!

oiami:
Added comment and squashed :)

oalders:
@oiami++

PR #1457, opened 2015-02-15
add explicit data: URI scheme support for use in images

zmughal:
This addresses the issue <https://github.com/CPAN-API/metacpan-web/issues/935>.

| Page | Before | After |
|--------|----------|--------|
| [Test::podimage](https://metacpan.org/pod/Test::podimage) |   |  |
| [IPerl example](https://metacpan.org/pod/release/ZMUGHAL/Devel-IPerl-0.002/example/20150209_IPerl_display_demo.pod) |    |  |


oalders:
@zmughal Thanks very much for doing the digging and figuring this one out.  :)

The failing tests were from an unrelated problem, so we're ready to merge!

PR #1460, opened 2015-02-22
Fix links in new feed and on news page.

davorg:
There were a couple of problems with the links in the News RSS feeds.

1. As they just use the title as the fragment, they were generating invalid URLs (ones that contained spaces).
2. The URLs in the feed didn't actually go anywhere useful - as the news page didn't include the &lt;a name="..."&gt; tags required to match the links in the feed.

This patch fixed both of those problems. It removes spaces from the URLs in the feed and adds matching &lt;a name="..."&gt; tags on the news page.

**However** the only way I could find to add the &lt;a name="..."&gt; into the news markdown was to insert raw HTML into the markdown just before it gets processed by the template engine. But you have HTML autoescaping turned on, which means my HTML is converted to &amp;lt;a name="..."&amp;gt; - which doesn't work.

So if you want your news page links to work then you'll need to a) apply this patch (or something similar) and b) turn off the HTML autoescaping.

If you can think of another solution, I'd love to hear it.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Feed.pm
+++ b/lib/MetaCPAN/Web/Controller/Feed.pm
@@ -35,9 +35,11 @@ sub news : Chained('index') PathPart Args(0) {
 
         my %e;
         $e{name} = $str =~ s/\A(.+)$//m ? $1 : 'No title';
+        my $a_name = $e{name};
+        $a_name =~ s/\W+//g;
         $str =~ s/\A\s*-+//g;
         $e{date}   = $str =~ s/^Date:\s*(.*)$//m ? $1 : '2014-01-01T00:00:00';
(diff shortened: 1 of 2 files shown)


haarg:
Fixing the links in the feed seems good.

The new page already includes ids on the headings, and that is what should be used for link targets.  Adding more <a> tags is pointless.

davorg:
Unless I'm missing some new HTML5 cleverness, you can't target links at just any id on a page - it has to be a named 'a' tag.

haarg:
ids have been valid targets since at least html4

davorg:
Damn. You're right. How did I miss that.

Ok. I'll withdraw this PR and submit a better one.

PR #1463, opened 2015-02-26
do not display .pod versions of documentation

karenetheridge:
.pod files in the root level of a Foo-Bar-Baz distribution are installed into
lib/Foo/Bar/*.pod, which pollutes the installation, so we should not encourage
this practice.


oalders:
Thanks!

PR #1484, opened 2015-04-04
Added in a boiler plate for installing cpan modules

sunnypatel4141:
Added in the show install instructions as per ticket https://github.com/CPAN-API/metacpan-web/issues/1295 This is my first pull request on this repo. please point me in the right direction if I have done anything in an unconventional manor.

### DIFF
--- a/root/inc/module-install.html
+++ b/root/inc/module-install.html
@@ -1,0 +1,35 @@
+
+<script type="text/javascript">
+	function show_install_instructions() {
+		jQuery('#install_module').modal('show');
+	}
+</script>
+
+<div id="install_module" class="modal fade">
+	<div class="modal-dialog">
+      <div class="modal-content">
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+          <h4 class="modal-title">Module Install Instructions</h4>
+        </div>
+        <div class="modal-body">
+		<p>To install <%  module.documentation or module.module.0.name %>, simply copy and paste either of the commands in to your terminal</p>
+			<p>CPANminus</p>
+			<pre class="bash: js">
+				sudo cpanm install <%  module.documentation or module.module.0.name %>
+			</pre>
+			<p>CPAN shell</p>
+			<pre class="bash: js">
+				sudo perl -MCPAN -e shell
+				install <% module.documentation or module.module.0.name %>
+			</pre>
+		<p>For more information on module installation please visit <a href="http://www.cpan.org/modules/INSTALL.html">The detailed CPAN module installation guide</a></p>
+        </div>
+        <div class="modal-footer">
+          <a href="#" data-dismiss="modal" class="btn">Close</a>
+        </div>
+      </div>
+	</div>
+
+</div>
+
--- a/root/inc/module-install.html
+++ b/root/inc/module-install.html
@@ -1,10 +1,3 @@
-
-<script type="text/javascript">
-	function show_install_instructions() {
-		jQuery('#install_module').modal('show');
-	}
-</script>
-
 <div id="install_module" class="modal fade">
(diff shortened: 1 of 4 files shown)


oalders:
@sunnypatel4141 thanks very much for this!  We'll make a few comments and this should be ready to merge very soon.

root/inc/module-install.html
```diff
@@ -0,0 +1,35 @@
...
+      <div class="modal-content">
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+          <h4 class="modal-title">Module Install Instructions</h4>
+        </div>
+        <div class="modal-body">
+		<p>To install <%  module.documentation or module.module.0.name %>, simply copy and paste either of the commands in to your terminal</p>
+			<p>CPANminus</p>
+			<pre class="bash: js">
+				sudo cpanm install <%  module.documentation or module.module.0.name %>
```
oalders: I don't think we want to default to installing modules via `sudo`, since that would mess with the system Perl.  We could either drop the `sudo` completely or offer it as an alternative with the caveat that you really should only do this if you know what you're doing.

root/inc/module-install.html
```diff
@@ -0,0 +1,35 @@
+
+<script type="text/javascript">
```
oalders: I was going to suggest that we don't use inline scripts and to put this logic into `./root/static/js/cpan.js`, but I see that we do have other templates with `<script>` tags.  I would think moving it to cpan.js would be the best way to handle this. @rwstauner, do you have an opinion?

root/inc/release-tools.html
```diff
@@ -21,6 +21,11 @@
     </a>
 </li>
 <li>
+	<a href="#" onclick="show_install_instructions(); return true;">
```
oalders: I think the way we normally handle this is to use jQuery to bind a click event to an element.  You can grep for `click` in `cpan.js` to see examples of how we do this.

oalders:
Just had a couple of quick comments.  Should be easy to sort out.  Thanks very much for your efforts on this! :+1:

root/inc/module-install.html
haarg: +1

root/inc/module-install.html
haarg: +1

root/inc/release-tools.html
haarg: Also, this should be a <button> element, since it isn't really a link.  Applying the class btn-link should make it look like the other links.

sunnypatel4141:
@oalders and @haarg I have taken your suggestions and implemented them. please take a look and let me know if there are any further changes. :+1:

oalders:
@sunnypatel4141 thanks for making those changes.  When I test it in Chrome, there's an overlay over the whole window and I can't close the modal:

screenshot 2015-04-08 22 29 43

Firefox and Safari look great.

sunnypatel4141:
@oalders I could not fully replicate this at first but this seems like an issue as mentioned here http://stackoverflow.com/questions/10636667/bootstrap-modal-appearing-under-background so as per the suggestion on the answer I moved the dialog out at the bottom of `pod.html` I could not think of a common place to put it. As we only want this to appear on the pod page.

oalders:
@sunnypatel4141 Thanks so much for the fixes.  I squashed the commits into 1 commit and pushed it directly to master.  Am deploying right now.

haarg:
A few more issues:
* hard tabs in html and js files
* CPANminus should be all lowercase, and maybe linked
* cpanm command is wrong, 'install' shouldn't be there
* extraneous newlines at end of `<pre>` blocks
* useless classes on `<pre>` blocks
* dialog code not included on release pages, yet sidebar button still exists

PR #1490, opened 2015-04-09
Using https for items at /feed/recent

bessarabov:
See https://github.com/CPAN-API/metacpan-web/issues/1488


oalders:
:+1:

bessarabov:
Thank you!

PR #1495, opened 2015-04-13
Feed url as https://

dolmen:
Fix the feed URL and URLs in the feed itself to always use `https://` instead of `http://`.


oalders:
:+1:

oalders:
Thanks, @dolmen!

PR #1533, opened 2015-05-23
randomly promote Perl events

szabgab:
It would be nice if we could use MetaCPAN to tell people about Perl-related events.
This patch could be used for experimentation with the idea.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Root.pm
+++ b/lib/MetaCPAN/Web/Controller/Root.pm
@@ -1,6 +1,9 @@
 package MetaCPAN::Web::Controller::Root;
 use Moose;
 use namespace::autoclean;
+use JSON::MaybeXS qw/decode_json/;
+use Path::Tiny qw/path/;
+use DateTime ();
 
 BEGIN { extends 'MetaCPAN::Web::Controller' }
 
@@ -102,6 +105,15 @@ sub end : ActionClass('RenderView') {
         . $c->config->{consumer_key};
 
     $c->stash->{site_alert_message} = $c->config->{site_alert_message};
+
+    my $file = $c->config->{home} . '/events.json';
+    eval {
+        my $all_ads = decode_json( path($file)->slurp_utf8 );
+        my $date    = DateTime->now->ymd;
+        my @ads     = grep { $_->{date} gt $date } @$all_ads;
+        my $ad      = $ads[ int rand( scalar @ads ) ];
+        $c->stash->{adserver} = $ad->{text};
+    };
 
     $c->fastly_magic();
 
--- a/root/wrapper.html
+++ b/root/wrapper.html
@@ -185,6 +185,7 @@
                                 <span class="input-group-btn">
                                 <button class="btn search-btn" type="submit">Search<% IF req.uri.path == '/mirrors' %> Mirrors<% END %></button>
                                 </span>
+                                <% adserver | none %>
                                 <small class="smaller">
                                 <% IF total.defined; IF page > 1 %>Page <% page %> of <% END; IF total == 999; 'more than '; END; total | format_number %>
                                 result<% IF total != 1; 's'; END %> (<% took / 1000 %> seconds)<% ELSE; "&nbsp;" | none; END %>
--- a/events.json
+++ b/events.json
@@ -1,0 +1,14 @@
+[
+   {
+     "text": "<a href=\"http://www.yapcna.org/yn2015/\">YAPC::NA</a> June 8-10, Salt Lake City, Utah",
+     "date" : "2015-06-08"
+   },
+   {
+     "text": "<a href=\"http://act.yapc.eu/ye2015/\">YAPC::EU</a> Sept 2-4, Granada, Spain",
+     "date": "2015-09-02"
+   },
+   {
+     "text": "<a href=\"http://yapcasia.org/2015/\">YAPC::Asia</a> Aug 20-22, Tokyo, Japan",
+     "date": "2015-08-20"
+   }
+]


ranguard:
It would be good if people could to submit news stories to perlnews.org - which appears on perl.org home page as well

szabgab:
I don't understand how perlnews.org and perl.org are related to this patch.

ranguard:
Well if you are trying to get the news out then that's a good channel, I didn't realise you'd already done so (I've approved your post and will figure out why I wasn't alerted to it)

szabgab:
I don't have the numbers for perlnews.org or perl.org, but I think a lot more people see metacpan.org.
(according to Alexa perl.org gets slightly more than blogs.perl.org that gets 50K visits/month while metacpan gets about 190K visits/month.)

If it was up to me, I'd use all the channels.

oalders:
I like this idea.  I think MetaCPAN is well placed to get the word out about things that are happening in the Perl world.  I do have a couple of points I think we should address here, though.

1) This adds more logic to the templates.  My strong preference would be to keep as much logic out of the views as possible.  (Please forget everything else you've ever seen in the MetaCPAN views).  There's a precedent there that makes maintenances (IMHO) harder than it should be.

2) These events have no expiration date.  So, we'd essentially need to pay attention to when these events are over and than manually remove them, push and merge the code and deploy.  If we just had the expiration then stale data is a non-issue.

3) I'm not sure I want these events right inside of the template.  I'm willing to be convinced that I'm wrong, but I'd rather see these in a JSON file that gets slurped by a method in a Model class which then produces some object that the ad can be generated from.  Or maybe this could all happen in JS, which would push it to the client side and reduce a small bit of work that would happen on every page load.

I don't want to make this overly complicated, but I think we should consider this stuff before deploying.

shadowcat-mst:
wasn't there a javascript widget thing that showed perl events? why can't metacpan just call that?

szabgab:
@shadowcat-mst probably refers to http://pcas.szabgab.com/ that I have basically discontinued.

@oalders 
1) I agree these should not be in the template. I implemented them this way because this was the easiest that hopefully allows discussion here, and if deployed we can see if people react to it positively or not. Then if you want to keep it then it can get a more robust implementation.

2) Agree. If this makes you accept the implementation as an experiment as in 1) I can add the expiration dates.

3) I am not sure I understood this. I would not put it in JavaScript. I'd probably have <% ad %> in the template and have the model supply the current ad. The list of ads could be in a JSON file as you suggested.

oalders:
@szabgab If you could move the logic out of the template, put the data into JSON and add an expiration, then we can merge this.  I guess the Perl ad server would have been ideal for this, but if it was underused I can understand why you would retire it.

szabgab:
Replaced the JavaScript version by a Perl version using a JSON file as the source including end-dates.

root/wrapper.html
```diff
@@ -179,6 +179,7 @@
                                 <span class="input-group-btn">
                                 <button class="btn search-btn" type="submit">Search<% IF req.uri.path == '/mirrors' %> Mirrors<% END %></button>
                                 </span>
+                                <% adserver | none %>
```
oalders: What's the purposes of `none` here?

oalders:
@szabgab Thanks! Could you rebase to get past the merge conflict?

root/wrapper.html
szabgab: Without that the HTML would be escaped.

szabgab:
rebased

oalders:
I apologize for taking _over one month_ to get back to this.  Here's what I see:

screenshot 2015-07-08 22 35 29

I want to get some feedback about the styling and placement before we go live with this.  @rwstauner? @ranguard? @haarg?

It's a bit subtle where it currently is.  We could use this same technique to put in small ads for cool, new modules as well, so let's think about how best to display this information.

oalders:
And don't worry about the rebase.  I'll handle that myself when it's ready to merge.

haarg:
With the location shown, it make it look like it's somehow related to the page content.  I don't really like it.

lib/MetaCPAN/Web/Controller/Root.pm
```diff
@@ -103,6 +106,15 @@ sub end : ActionClass('RenderView') {
 
     $c->stash->{site_alert_message} = $c->config->{site_alert_message};
 
+    my $file = $c->config->{home} . '/events.json';
+    eval {
+        my $all_ads = decode_json( path($file)->slurp_utf8 );
```
rwstauner: This should be `slurp` (or `slurp_raw`).  `decode_json` expects UTF-8 bytes, not a character string.

rwstauner:
I agree, it looks as though the "732 results" is related to the event... is that how many people are registered? :)

kentfredric:
It occurs to me that in the screenshot above, that there is a copious amount of unused whitespace on the left hand sidebar in search results that could be utilized here.

Similarly, there's plenty of space on the left of the "Recent" feed.

Given "events" are time-temporal, they seem to be associated loosely with the concept of "Recent Modules", so a "Recent" feed and a "Recent Events" don't seem too  disparate.

There's also the metacpan front page that is quite swimming in whitespace where events could be shown.

There's additionally a bit of space in the grey bar at the top of the screen that could be used to store a single-line event ticker. 

The trick is having things visible enough to see, but not so visible they affect workflow.

> 2) These events have no expiration date. So, we'd essentially need to pay attention to when these events are over and than manually remove them, push and merge the code and deploy. If we just had the expiration then stale data is a non-issue.

Surely, even if the event itself had no formal expiration, you could enter the data with a "Display-Until: " property. 

Though reading the messages, I can't tell what you mean by "These events have no expiration date", I can't tell if that means "The data source(re:perl news) that the events are comming from does not advise when the event will terminate" or "The event itself will never terminate"

oalders:
How about moving the event notification into something like `<div class="alert alert-info" role="alert">...</div>`?  That would give it a blue background and set it off from the other information.

oalders:
@kentfredric I just meant that this PR as it stands doesn't have a way of _not_ displaying events after some arbitrary date.  What I wanted to avoid is having to merge code on a certain date in order to stop some event from appearing.

oalders:
@szabgab thanks very much for this.  I'm going to close this for now as there doesn't seem to be an activity.  We can happily re-open it if someone wants to finish this off.

PR #1537, opened 2015-05-27
Leo/switch css js to cdn friendly

ranguard:
Move js and css to fixed URL's (e.g. no md5... so works with cached pages), move lessc to server side and dynamic (also faster for development IMHO)

We'll delete fastly cache when we do a deploy (fastly setup to cache for 30 days), and users can cache for an hour

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -71,6 +71,7 @@ requires 'Plack::Test';
 requires 'Regexp::Common';
 requires 'Regexp::Common::time';
 requires 'Starman', '>= 0.4008';
+requires 'String::BOM', '0.3';
 requires 'Template::Alloy';
 requires 'Template::Plugin::DateTime';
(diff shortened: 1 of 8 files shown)


dolmen:
@oalders raised on IRC that this patch adds File::Slurp as a dependency.
But this is a really soft dependency: File::Slurp is used by String::BOM only for its testsuite. I've filled [RT 104739](https://rt.cpan.org/Ticket/Display.html?id=104739).

haarg:
As discussed on IRC, I'd prefer to use a different approach, still using hashed paths, so that the files can be cached indefinitely by browsers.  I have the code partly done, and I'm going to revisit it over the next couple days.

ranguard:
Closing as @haarg has a better solution he is going to work on

PR #1546, opened 2015-06-06
Leo/caching 1

ranguard:
First phase of setting cache headers.

Cache can be cleared with the bin/purge.pl script

### DIFF
--- a/lib/MetaCPAN/Web/Controller/About.pm
+++ b/lib/MetaCPAN/Web/Controller/About.pm
@@ -5,6 +5,16 @@ use Format::Human::Bytes;
 
 BEGIN { extends 'MetaCPAN::Web::Controller' }
 
+sub auto : Private {
+    my ( $self, $c ) = @_;
+
+    $c->add_surrogate_key('about');
+    $c->res->header(
+        'Cache-Control' => 'max-age=' . $c->cdn_times->{one_day} );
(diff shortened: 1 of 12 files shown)


lib/MetaCPAN/Web/Controller/About.pm
```diff
@@ -5,6 +5,16 @@ use Format::Human::Bytes;
 
 BEGIN { extends 'MetaCPAN::Web::Controller' }
 
+sub auto : Private {
+    my ( $self, $c ) = @_;
+
+    $c->add_surrogate_key('about');
+    $c->res->header(
+        'Cache-Control' => 'max-age=' . $c->cdn_times->{one_day} );
+    $c->cdn_cache_ttl( $c->cdn_times->{one_day} );
```
oalders: This all looks good to me.  Would be nice if the role had a cdn_cache_control() method similar to the cdn_cache_ttl() method.  Would eliminate some repetition in the controllers.  Not a dealbreaker, though.

oalders:
Looks good to merge.  I won't merge right now as I'm not around to fix any deployment issues.

lib/MetaCPAN/Web/Controller/About.pm
ranguard: I wondered about that, but 'Cache-Control' isn't specifically cdn functionality, it's just another header, that said I might review it again some time once I've done some more complex examples

PR #1548, opened 2015-06-16
use Constructor.io autocomplete for author search

danmcc:
This adds author search with autocomplete.


oalders:
The failing test was just an issue with tidying the JS.

oalders:
@rwstauner, @haarg did either of you want to review the JS and CSS changes?

root/static/js/cpan.js
```diff
@@ -183,41 +183,81 @@ $(document).ready(function() {
...
+            }
+        });
+    });
+
+    search_input.bind('authors_autocomplete', function() {
+        $(this).constructorAutocomplete({
+            key: 'bQssESEa8XUBKzSUjarO',
+            directResults: true,
+            maxHeight: 400,
+            transformResult: function(response) {
```
rwstauner: Is this actually used?  It looks like it's there to remove a search operator, but when I tried it locally I never saw anything that it would be removing.

rwstauner:
If the external script fails to load and the author button is selected, there is no autocomplete.  Switching back to module works fine, so I think that works as well as could be expected.

rwstauner:
Unfortunately this suffers from an issue we've gotten complaints about in the past with the module autocomplete: If you happen to have your mouse a little below the search box, type some characters, and press enter, instead of actually searching for your term, you go straight to the autocomplete item that was under the mouse, regardless of whether you ever moved the mouse or intended to choose that item.  In firefox this happens all the time.  In chrome, if i move the mouse once while over the autocomplete dropdown, and then change the search term entirely, i get the same (undesired) behavior.

You can see how we dealt with this in the lines of code directly following your addition (5db6e9529c169a242a931a77aa964fdb10a60cac).

Is that something that can be fixed?

root/static/js/cpan.js
danmcc: Is the transformResult() function being used, you mean?  Yes, we're doing some trickery to make the PAUSE IDs searchable but not display them as the primary search term.  So we're indexing the string "[PAUSE ID]: [Full name]" so that any match on that string returns results.  But then we're removing everything in front of the colon here so that the full name appears as the primary search term, and the PAUSE ID is the "description".

danmcc:
@rwstauner Great point about the mouseover and autocomplete.  I've moved your fix into the bind() events so that it gets re-enabled if the user switches back and forth between modules and authors.

root/static/js/cpan.js
rwstauner: I'll take your word for it, but I asked because I did a `console.log` on the value before it was transformed (out of curiosity) and i didn't see any values that would get transformed, that's why i asked if it might actually get used.

rwstauner:
Oh, nice!  Thanks for thinking that through.  I hadn't tested it diligently to see if it would need to be re-enabled after switching but i bet you're right.

rwstauner:
LGTM.  Are we ready to merge this?

danmcc:
We're all set on our side!

oalders:
:+1: I'll merge and deploy tomorrow unless somebody else beats me to it. :)

rwstauner:
deployed. Thanks!

danmcc:
Thanks, guys!  If there's anything we can do along the way to make it better, please let us know.

oalders:
@danmcc thanks for all of your hard work on this.  :)  I'm just noticing that the options seem to be restricted to 5 results.  For instance a search for "randy" doesn't include @rwstauner.  Is it possible to add more results or to make the list scrollable?

danmcc:
@oalders All set!  We now show 10 results.  Let me know what ya think.  I may tweak the styling a little so that the branding at the bottom shows up no matter what, but will get to that later.

oalders:
@danmcc Thanks! The 10 results seem better. I do see that the branding disappears if there are too many results.  Seems like a good idea to fix that. :)

PR #1549, opened 2015-06-16
point links to the API to the up-to-date location

### LINKED ISSUE #1513: "API" link in footer points to an obsolete link
In the footer of every MetaCPAN.org page, the "API" word links to https://github.com/CPAN-API/cpan-api/wiki/API-docs
However that page says:
> Note: Under Construction
> The contents of this page have been moved [into the repo](https://github.com/CPAN-API/cpan-api/blob/master/docs/API-docs.md)

So, the link should point instead directly to the current location.

book:
fixes #1513

### DIFF
--- a/root/about/faq.html
+++ b/root/about/faq.html
@@ -64,7 +64,7 @@ See our [missing module](/about/missing_modules) page.
 
 ## Where can I find the API docs?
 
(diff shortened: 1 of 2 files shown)


rwstauner:
Thanks!

PR #1552, opened 2015-06-21
fix #1547 added in code to release

### LINKED ISSUE #1547: "Install instructions" link doesn't work
This link on any page doesn't seem to do anything (OS X, Chrome and Safari)...

log-report-lexicon-1 04_-_log__report_translation_table_management_-_metacpan org

Didn't see any errors in console, just nothing.

Used to work, from https://github.com/CPAN-API/metacpan-web/pull/1484/files

sunnypatel4141:
Fixed the install template issue for release

### DIFF
--- a/root/inc/module-install.html
+++ b/root/inc/module-install.html
@@ -6,14 +6,14 @@
                 <h4 class="modal-title">Module Install Instructions</h4>
             </div>
             <div class="modal-body">
-                <p>To install <%  module.documentation or module.module.0.name %>, simply copy and paste either of the commands in to your terminal</p>
+                <p>To install <%  module.documentation or module.module.0.name or module.package %>, simply copy and paste either of the commands in to your terminal</p>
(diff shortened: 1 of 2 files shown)


rwstauner:
The cpan clients don't install by dist name, so this isn't right.
I would use `release.main_module`.
It's a newer field, so it's not available on all past releases until a full reindex, but it should actually work, since the field was designed for this sort of purpose: give me the most representative module of the dist (and the cpan clients require a module name to install).

rwstauner:
The clients can install by "$author/$archive", however, so that would be a more suitable fallback:
`release.main_module or (release.author _ "/" _ release.archive)`

oalders:
@sunnypatel4141 Did you see the comments on your pull request?

oalders:
@sunnypatel4141 thanks for this.  I'm sorry -- I didn't notice until just now that you had pushed that last commit.

PR #1560, opened 2015-07-14
work around Firefox bug where localStorage can throw an exception

### LINKED ISSUE #1558: lots of javascript broken if localStorage is blocked
If localStorage is blocked for some reason, attempting to use it will throw an exception.  This can prevent a bunch of the other UI code from running, even though our localStorage usage is all optional.

haarg:
If the user has cookies set to ask every time, accessing localStorage
will throw an exception.  Work around this by shim that just stores the
data for the current page view.

Fixes #1558


oalders:
:+1:

dolmen:
:clap:

PR #1566, opened 2015-07-25
PSGI and Asset rewrites

haarg:
This is a refactoring of the PSGI file, as well as a different way of handling css/js/less files.

The psgi file has been refactored to mainly use Plack::Builder, as it makes it much more clear.  The static file handling has been factored out into a separate middleware, which allows it to be used separately from the main app.  In a later commit, I've added a separate app for handling the static files that proxies to the main app.  This is the setup I use locally for development, as it allows a single process server to handle the main code, while an event based server hand handle the numerous static files that are needed.

The asset handling has been adjusted to build the processed output at server startup and save it to a static file.  The directory with this file is then served as a normal directory.  This means old copies of the css will continue to be accessible until they get cleaned up.  The files are also given a 1 year cache expiration.


oalders:
Thanks for putting this together.  I'm not concerned about the tidying failures, but there a some others that need some tweaks:

```
t/fastly_headers.t                     (Wstat: 256 Tests: 4 Failed: 1)
  Failed test:  4
  Non-zero exit status: 1
t/plack/mcless.t                       (Wstat: 512 Tests: 0 Failed: 0)
  Non-zero exit status: 2
  Parse errors: No plan found in TAP output
```

ranguard:
my only issue (from skimming the code) with this is the 'cache_dir' should be a repo or something, so that it can be shared across machines, either we care about old versions (so they should be available on all boxes, freshly installed or just made production), or we don't, this doesn't quiet fit either.

The 'cache_dir' could be a git repo with a cron script to push it up to github or something, and I can then get puppet to pull it down each run, brain not fully in gear at the moment, so maybe a better approach

haarg:
You have a good point about the availability of the old files, but I'm not sure it's really a problem worth solving.

The current code we're using sets needlessly short cache timeouts on the CSS files, but no expiry on module pages etc.  This means browsers will cache module pages indefinitely, but CSS for only a short time.  It should be exactly the opposite of that.  Part of the changes I've made here sets a year long timeout on the CSS and JS files.  They have unique URLs, so there's no reason to ever have them expire.  And since Fastly will be caching the files, it should continue to work even when we are load balancing.  We also need to fix the module page caching, but that can be done separately.

ranguard:
We should never rely on Fastly to have our files in cache, sometimes they will, sometimes they won't (depending on MANY factors, but most obvious of which being us sending a full purge for some reason, which currently is any HTML change requires it, because we haven't got the surrogate keys setup for everything, as you say). - That said meh, don't feel strongly either way

haarg:
We aren't really relying on Fastly.  We're relying on having better cache settings on the CSS/JS files, so they will be kept longer than the pages referencing them.  And then Fastly and the files left on disk are two backup mechanisms just in case.

We could configure Fastly to try other backends if it gets a 404 on the asset files.

ranguard:
I just mean't "since Fastly will be caching the files, it should continue to work even when we are load balancing" isn't an ideal design decision, and getting Fastly to use different backends would be more work than just putting the contents into a repo. All of this is almost academic, as we don't switch backends that often in any case, and even if we do it's only old browser cached pages that would be affected.

oalders:
I'll leave this merge up to @ranguard and @haarg, but there is still the matter of the failing tests that need to be fixed.

ranguard:
@haarg over to you, go for it :)

lib/MetaCPAN/Middleware/Static.pm
```diff
@@ -0,0 +1,106 @@
...
+
+    my @less_files = ('/static/less/style.less');
+
+    builder {
+        if ( !$dev_mode ) {
+            enable 'Assets::FileCached' => (
+                files => [ map "root$_", @js_files ],
+
+                # need js compressor
+                #filter => sub { },
```
oalders: Is there something that still needs to be added here?

haarg: I'll push a commit adding JavaScript::Minifier::XS, which is what was used before.  I'd like to start using closure compiler, but that requires more things to be installed.

haarg:
I've updated this to include Javascript minification.

My one remaining concern is that I've included a script for running a dev server in the way that I do.  It doesn't use carton, so it's a little out of the norm for how most of our docs etc are done.  Possibly I should just add some docs to the script, or maybe it could be adjusted to work more in line with how others work.  Or maybe it should just be left out entirely.  It also relies on some modules that aren't included in the cpanfile.

Issue #1554, opened 2015-07-07
New radio buttons unclear


kentfredric:
The purpose of the radiobuttons listed "module" and "author" are very vague and eat up parts of the page that are useful.

If you don't know what they're for, then you get weird results:

![sn154](https://cloud.githubusercontent.com/assets/44790/8540159/07a1c3ec-24d7-11e5-86fe-d7991ab88cb8.png)
![sn153](https://cloud.githubusercontent.com/assets/44790/8540160/07a8d0e2-24d7-11e5-9303-a1b8683d5406.png)

Not until I went to complain and saw the commit message did it become slightly clear:

![sn155](https://cloud.githubusercontent.com/assets/44790/8540182/398608be-24d7-11e5-99c5-055d70550428.png)

Only then did it become obvious how they worked.

ranguard:
I get this on Chrome on a mac:
![search_for__data__page__-_metacpan_org](https://cloud.githubusercontent.com/assets/60391/8541403/e07ef088-2485-11e5-9d58-ab0e0c564cfe.png)

oalders:
ping @danmcc

rwstauner:
Linux Chrome, also.

>   I guess I only looked at firefox. :\

oalders:
Looks fine on my OS X Chrome `Version 43.0.2357.130 (64-bit)`

danmcc:
Weird!  I'll take a peek.  I'm totally open to other UI implementations, too.  We could have a select list between the text field and the search button sort of like [Shutterstock](http://www.shutterstock.com/) does.

ranguard:
+1 for shutterstock drop down style selector

oalders:
+1 for Shutterstock style.  Didn't even know that was an option. :)

oalders:
@kentfredric care to comment on the UI in #1569 ?

ranguard:
No feedback received - closing

PR #1569, opened 2015-08-06
use dropdown for module/author search selection

danmcc:
This addresses issue #1554, at least on the homepage.  Take a peek and let me know what ya think.  If it looks good, I can put it on search result pages, too.  Feel free to edit the implementation as you see fit.  I'm not the world's best front-end developer, so am happy to have someone else improve on it.

### DIFF
--- a/root/home.html
+++ b/root/home.html
@@ -1,16 +1,39 @@
 <%- rss = "recent" %>
 <div align="center" class="home">
   <a href="/" class="big-logo" alt="meta::cpan"></a>
-  <form action="/search">
-    <input type="text" name="q" size="41" autofocus="autofocus" id="search-input" class="form-control home-search-input">
+  <form action="/search" class="form-search">
     <input type="hidden" name="size" id="size" value="20">
-    <input type="radio" class="search-radios" id="search_modules" name="search_type" value="modules" checked> <label for="search_modules" class="search-labels"> Modules</label>
-    <input type="radio" class="search-radios" id="search_authors" name="search_type" value="authors"> <label for="search_authors" class="search-labels"> Authors</label>
+    <div class="search-wrapper">
+      <div class="input-group">
+        <div class="form-control">
+          <input class="search-input" id="search-input" name="searchterm" type="text">
+        </div>
+        <div class="dropdown input-group-addon">
+          <div class="dropdown-wrapper">
+            <span class="dropdown-trigger" data-toggle="dropdown">Modules <span class="caret"></span></span>
+            <ul class="dropdown-menu">
+              <li>
+                <label for="radio-modules">
+                  <input class="visuallyhidden" id="radio-modules" type="radio" name="search_type" value="modules" checked="checked">
+                    Modules
+                </label>
+              </li>
+              <li>
+                <label for="radio-authors">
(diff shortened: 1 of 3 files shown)


oalders:
@danmcc, thanks for this!  Sorry for my delay in reviewing.

I think it's just the tidy test that is failing here.  That's easy for us to fix.  I've included a screen shot of the changes.  I think we may want to add some kind of colour to the dropdown.

screenshot 2015-08-19 09 39 35

Anyone have thoughts on the UI?

oalders:
@rwstauner, @haarg, @ranguard feedback?

rwstauner:
One problem i see is that the cursor changes on the wrapper, but the dropdown trigger is smaller, so the first few times i clicked, nothing happened.
screenshot - 08222015 - 10 08 24 am

rwstauner:
Another is that we lost some valuable attributes on the searchbox like `name="q"` and `autofocus`.

rwstauner:
Since this is only for autocomplete (it doesn't affect the search results)
i wonder if we should say "Autocomplete" somewhere (it would have fit before the radios, it could fit in the new button, or maybe just title attribute (hover text))...

kentfredric:
I guess some of the confusion in #1554 for me really, is due to the fact I can't think of another time I wanted a choice for an autocomplete

Autocomplete is supposed to be inherently a simple things that give a best first-pass without needing to think, and even providing a choice mechanism there just adds to confusion.

I'd be ok with the selection changing how the *main* search works, but it just doesn't, and that it doesn't is part of the confusion.

Instead of proving a choice of which completion to use, seeing we only have two, why can't we have the results of both in the same? 

snapshot5

Or maybe something inline that assumes you're already done typing and then switches the result drop down?

snapshot6

danmcc:
Sorry for the delay on this.  I finally got time to sit down and play with it a bit and create a mockup.  If we combine the two datasets, we could serve a single autocomplete as shown below.  (This would require moving the module autocomplete to Constructor.io, but I think that would offer a huge improvement in the results that are shown.)

What do you all think?

image

oalders:
Visually, that works for me.  My main concern would be around keeping the data up to date on your end.  What would your plan be for that?  Can we push to you or do you need to pull?

danmcc:
Yup, you can push to us.  We'd work together to integrate our [API](constructor.io/docs).  I can start poking around with it if you point me to where in the code it'd make sense to push to us.

oalders:
So, I think I'd like us to come back to this _after_ the QA hackathon, which is in about 3 weeks.  Once we've moved to Elasticsearch 2.0 we'll have a better idea of how our module search is broken (or fixed?)

However, it looks like we could use https://metacpan.org/pod/WebService::ConstructorIO to come up with the integrated author and module search suggestions.

@danmcc could you set us up with the account credentials to do this?  You can email me directly.  I'd be interested to see how easy this is.

danmcc:
Will do @oalders!  Our Perl API docs are [here](https://constructor.io/docs/?perl) if you'd like to take a gander.

mickeyn:
is this PR still relevant?

oalders:
I think with @haarg adding the author search to the dropdown it's no longer needed.  It's my fault for not getting to this.  I had the best of intentions, but at the end of the day, I just didn't get to it.  My apologies to @danmcc :(

PR #1583, opened 2015-08-23
Json view errors

rwstauner:
Ever since we added the `favorites_as_json` thing i've been seeing a 500 Error in my console about it.
I noticed this week that it is actually killing the Account menu for me.
I **am** logged in, but b/c the json call fails, the javascript shows me the "Sign in" menu, which is very confusing.

I noticed an error in the log:
```
[error] Caught exception in MetaCPAN::Web::View::JSON->process "encountered object 'true', but neither allow_blessed, convert_blessed nor all
ow_tags settings are enabled (or TO_JSON/FREEZE method missing) at (eval 891) line 151."
```

and tracked it down to the fact that my user profile has booleans in the "extra" data.

### DIFF
--- a/lib/MetaCPAN/Web/View/JSON.pm
+++ b/lib/MetaCPAN/Web/View/JSON.pm
@@ -1,8 +1,14 @@
 package MetaCPAN::Web::View::JSON;
 
 use Moose;
+use JSON::MaybeXS ();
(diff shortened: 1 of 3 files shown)


ranguard:
Nice spot!

PR #1586, opened 2015-08-25
Replace MetaCPAN Helper with MetaCPAN redirect

petrkle:
[Reviews](https://chrome.google.com/webstore/detail/metacpan-helper/aoioenbjpmccpkincghhjfmeceknpcnb/reviews) says that old chrome extension does not seem to work. Link to source code is [broken](https://github.com/worr/metacpan-ext).


oalders:
Thanks @petrkle.  :+1:

Issue #1588, opened 2015-08-29
Make the "whatsnew" section stand out more


oalders:
Aside from fixing the issue with Changesets which are too large, we need to set this apart a bit.  If I just put it into a well, it doesn't play terribly nicely with the right panel.

<img width="1225" alt="screenshot 2015-08-28 23 01 00" src="https://cloud.githubusercontent.com/assets/96205/9560265/ebd39dae-4dd8-11e5-8396-c2baf76db796.png">

haarg:
The overlap with the sidebar can be dealt with pretty easily by using a margin.

oalders:
Not being well versed in CSS, can you point me at where I'd be adding the margin?

PR #1599, opened 2015-10-11
Make what's new stand out with a well and margin.

bentglasstube:
Per the discussion for #1588, I have put the whatsnew section in a well and added a margin to prevent it from overlapping with the right panel.


bentglasstube:
I changed it to 200px but my vagrant image was not cooperating so I didn't see how that actually looked with the sidebar.

oalders:
Thanks!

PR #1613, opened 2015-11-07
Reduce id conflicts and other cleanups

haarg:
This tries to reduce the chance of having conflicting ids between pod content and the rest of the site.  Includes some other miscellaneous fixes and cleanups.


oalders:
:+1:

Issue #1640, opened 2016-01-29
Logged In Users Favourites fetched on every page request


kentfredric:
I was trying to diagnose why my page loads were taking so long, and I found some "fun" things.

Every page request fetches `https://metacpan.org/account/favorite/list_as_json`, logged in or not. 

And when diagnosing the problem using webpagetest.org ( as recommended by fastly ), this URI stuck out as being a significant slowdown, taking 500ms to query, ... and then returning 403.

> https://metacpan.org/source/RJBS/if-0.0606/Changes
> ![snapshot](http://www.webpagetest.org/waterfall.php?test=160129_ZR_K03&run=1&cached=0)

Upon performing this myself, I found that a lot of metadata is being fetched without an obvious application. Specifically, its strange to be seeing so much metadata about "who I am" and "my favourites" when I'm on a page unrelated to "me", and I'm not seeing how that data is useful except for the fact I have to be logged in _somehow_

kentfredric:
This behaviour appears to be one of the largest perfomance limiters for at least logged-out clients, as the 403 result can't be cached, but is still part of *every* request.

And so that nice big delay is incurred on every fetch, and blocks rendering the page.

Here's a test I did on a page I expected not to already be on the CDN from an Australian test machine connected at "Native" speed, which could very much be in the same cabinet as a fastly server.

http://www.webpagetest.org/result/160129_8V_M07/

There are 2 "jobs", and each job has 2 requests: The initial page load ( which is ~2.5 seconds ), and a second page load ( a refresh , and takes ~0.7 seconds ). 

Of this refresh, 0.5 of those 0.7 seconds are consumed by the null `list_as_json` call.

There's a bunch of other ripe targets that could be improved on greatly, but this seems the largest, most obvious one, and the one with the least explanation for why it has to be there.

Other candidates for "bad" content that is uncached and potentially blocking page render:

```
https://cnstrc.com/js/ac.js?_=1454069053076 # no max age, no expires
https://api.coderwall.com/rjbs/endorsecount.png # expires in 60 seconds
https://secure.gravatar.com/avatar/cec6b9b88295ead26c65f8747a76561b?s=130&d=identicon # expires in 5 minutes
```
These also seem to have reasonably short caching times
```
7.3 minutes - https://metacpan.org/static/icons/metacpan-icon.png
26.0 minutes - https://metacpan.org/static/icons/favicon.ico
[...]

PR #1642, opened 2016-01-31
Cache assets at fastly for a year #1640 - we can purge "assets" to clear

ranguard:


oalders:
@ranguard could you rebase?

oalders:
I just rebased and force-pushed.

PR #1637, opened 2016-01-17
fixed whitespace if rating was non-integer

mephinet:
@haarg reported that #1633 introduced a layout issue if the rating of the module was non-integer. This pull request fixes the whitespace.

PR #1659, opened 2016-03-20
fixed whitespace if rating was non-integer

oalders:
Rebased version of #1637


haarg:
The `<span class="rating-<% stars * 10 %>">` tag is unclosed in this.  It should be closed immediately after it is opened.

oalders:
@haarg is this better?

PR #1662, opened 2016-03-30
No backpans

oalders:
This is a rebased version of https://github.com/CPAN-API/metacpan-web/pull/1609


oalders:
@mickeyn this does not fix the actual problem, but it does take care of front end issues.  We should revert this once we have a fix in the back end, but for now I think it's fine to merge.

PR #1667, opened 2016-04-06
[WIP] some hacks to allow working against ES 2.3 index

mickeyn:
Some of the hacks used (except for local configuration files) to allow working with ES 2.3 API
I assume this is far from being complete and this PR is only meant for the discussion and change reference for actual future changes.

### DIFF
--- a/lib/MetaCPAN/Sitemap.pm
+++ b/lib/MetaCPAN/Sitemap.pm
@@ -8,7 +8,8 @@ use warnings;
 use autodie;
 
 use Carp;
-use ElasticSearch;
+use Search::Elasticsearch;
+use ElasticSearch::SearchBuilder;
 use File::Spec;
 use MetaCPAN::Web::Types qw( HashRef Int Str );
 use Moose;
@@ -51,11 +52,14 @@ sub process {
     # actually exist and b) the directory itself is writeable.
 
     # Get started. Create the ES object and the scrolled search object.
-    my $es = ElasticSearch->new(
-        servers    => 'api.metacpan.org',
-        no_refresh => 1,
+    my $es = Search::Elasticsearch->new(
+        #        nodes            => ['api.metacpan.org'],
+        nodes            => ['metacpan-01.ams4.prod.booking.com'],
+        cxn_pool         => 'Static::NoPing',
+        send_get_body_as => 'POST',
     );
-    defined $es or croak "Unable to create ElasticSearch: $!";
+
+    my $field_name = $self->field_name;
 
     # Start off with standard search parameters ..
 
@@ -63,7 +67,7 @@ sub process {
         index  => 'v0',
         size   => 5000,
         type   => $self->object_type,
-        fields => [ $self->field_name ],
+        fields => [$field_name],
     );
 
     # ..and augment them if necesary.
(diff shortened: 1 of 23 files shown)


mickeyn:
running the test suite reveals more issues, and I'm attacking one by one... feel free to comment / patch

ranguard:
Very quick skim... looking good :)

lib/MetaCPAN/Sitemap.pm
```diff
@@ -118,4 +121,6 @@ sub process {
     return;
 }
 
+__PACKAGE__->meta->make_immutable;
```
oalders: :+1:

lib/MetaCPAN/Web/Model/API/Release.pm
```diff
@@ -198,17 +198,20 @@ sub modules {
             # Sort by documentation name; if there isn't one, sort by path.
             sort => [ 'documentation', 'path' ],
 
-            # Get indexed and authorized from _source to work around ES bug:
-            # https://github.com/CPAN-API/metacpan-web/issues/881
-            # https://github.com/elasticsearch/elasticsearch/issues/2551
```
oalders: Wondering why the comments are removed.  Are they no longer relevant?

lib/MetaCPAN/Web/Controller/Author.pm
```diff
@@ -50,15 +51,15 @@ sub index : Chained('root') PathPart('') Args(0) {
     my ( $author, $data ) = ( $author_cv->recv, $releases_cv->recv );
     $c->detach('/not_found') unless ( $author->{pauseid} );
 
-    my $faves_cv = $c->model('API::Favorite')->by_user( $author->{user} );
+    my $faves_cv = $c->model('API::Favorite')->by_user( $author->{pauseid} );
```
oalders: Wondering about this change (and the other similar one).  Non-authors can also use the `++` functionality, so we need to make sure we're not breaking anything for them.

lib/MetaCPAN/Web/Model/API/Release.pm
mickeyn: both issues indicate a fix at ES 0.90. seemed not relevant to me.

lib/MetaCPAN/Web/Controller/Author.pm
mickeyn: I didn't find the `user` entry in `$author`. it was a hack to make it work - but I agree it may have been a wrong choice.
Will have another look.

oalders: Ah, ok.  It should be there.  It's there on production.  http://api.metacpan.org/author/OALDERS?fields=user

lib/MetaCPAN/Web/Model/API/Release.pm
oalders: I see you're right.  Thanks!

lib/MetaCPAN/Web/Controller/Author.pm
mickeyn: reverted the change. thanks @oalders ;)

PR #1674, opened 2016-04-15
updated cpanfile

mickeyn:
Updating cpanfile to use `Search::Elasticsearch`.
There's still use of `ElasticSearch::SearchBuilder` in the code, I'll look into that one next.


mickeyn:
OK, I see that `ElasticSearch::SearchBuilder` is only used in `MetaCPAN::Sitemap` in one case of the `releases.xml.gz` where the filter is hardcoded as:
```
filter => { status => 'latest' }
```
and this module is transforming this into:
```
{
        query   {
            match   {
                status   "latest"
            }
        }
}
```

@oalders: we can leave it as it works, but it seems a bit of an overkill, wouldn't replacing:
```
$search_parameters{'body'}
            = ElasticSearch::SearchBuilder->new->query( $self->filter );
```
with:
```
$search_parameters{'body'}{'query'}{'match'} = $self->filter;
```
work just the same and we remove a module dependency completely (at least until we need it)? my worry is that if we use it for anything more complex we might run into incompatibilities in ES2.x (`ElasticSearch::SearchBuilder` is from 2013)

ranguard:
@mickey - go for it - we shoudn't use ElasticSearch::SearchBuilder  at all
(even clinton doesn't recommend it now).

p.s. I might not be able to do much tonight/tomorrow - will try squeeze in
deploying!

mickeyn:
done :)

oalders:
Thanks!

Issue #1671, opened 2016-04-12
The changes box should include all the changes from the latest stable release.


salva:
Currently, when a module page is visited, it includes the set of changes for the last release, but for modules that have intermediate development releases that doesn't work well.

It would be better to show all the changes since the previous stable.

For instance, the page for [Net::SSH2 0.58](https://metacpan.org/release/SALVA/Net-SSH2-0.58) shows...

    Changes for version 0.58
         - re-released as stable

It would be much interesting to show instead the following:

    Changes for version 0.58
        - re-released as stable
    Changes for version 0.57_03
        - remove MYMETA.* from distribution
        - workaround bug in libssh2_agent_disconnect
    Changes for version 0.57_02
        - Module::Install::CheckLib was missing
    Changes for version 0.57_01
        - reimplement scp_put and scp_get methods fixing several issues
        - add support for Mac Homebrew and its keg-only OpenSSL

oalders:
@haarg  is this possible?

haarg:
This has been implemented.

PR #1680, opened 2016-04-23
Show changelogs since last stable

### LINKED ISSUE #1624: Change logs not parsed properly for TRIAL releases
The Changes section on `/release/...` doesn't show up for `-TRIAL` releases.  Example: https://metacpan.org/release/TSIBLEY/App-RecordStream-4.0.10-TRIAL

Parsing of changelogs also seems busted if `-TRIAL` is involved in the changelog release header, as you can see in the parse of the 4.0.11 changelog which includes 4.0.10-TRIAL: https://metacpan.org/release/TSIBLEY/App-RecordStream-4.0.11

haarg:
This causes metacpan to show all changelog entries since the last stable release.  Also fixes the handling of -TRIAL releases, and has a cleanup of the show/hide mechanism.

Fixes #1624 and #1671.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Release.pm
+++ b/lib/MetaCPAN/Web/Controller/Release.pm
@@ -131,8 +131,14 @@ sub view : Private {
                 qw( mark_unauthorized_releases )
         ),
 
-        ( $changes ? ( last_version_changes => $changes ) : () )
-
+        (
+            @$changes
+            ? (
+                last_version_changes => $changes->[0],
+                changelogs           => $changes,
+                )
+            : ()
+        )
     );
 }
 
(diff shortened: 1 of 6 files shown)


oalders:
I'd like to merge this, but I also don't want to create any merge conflicts with the `mi/es2` branch.  I think we should just wait for a few days until we have an idea of how long that other branch is going to live.

haarg:
This branch currently merges into mi/es2 without any conflicts.

oalders:
Perfect.  Can you create another pull request for `mi/es2` and we'll merge this into that?  We're (hopefully) past the point of backporting any changes from master into that branch.  I'd like to consider all the master branches basically frozen, unless there's a compelling argument otherwise.

haarg:
Taking another minute to think about this, I'd say it should use the MetaCPAN index data to choose if a given release is a beta or not.  However, I'd still advocate for merging this as is, leaving that for a future improvement.

PR #1687, opened 2016-05-05
Drop recent/log - not used much

ranguard:
Would also need to use a different backend mechanism as we are heading towards using the queue which is distributed


ranguard:
21:01 < haarg> ranguard: i'd suggest merging master before making that change ^
21:04 < ranguard> oalders: ^^ :)
21:05 < haarg> master should be cleanly mergable into mi/es2, but that change will cause a conflict
21:07 < ranguard> haarg: just to confirm that's currently to be merged back to mi/es2... not master
21:08 < haarg> yes.  i'm saying that currently master can be merged to mi/es2.  with that change, there
               will be conflicts to resolve.
21:08 < haarg> i'd suggest going that merge now, then making your changes on top of that
21:09 < ranguard> I'll stick a not on the PR - cheers

PR #1693, opened 2016-05-14
Have both CPAN author directory link and BackPAN author directory link

neilb:
Update author page to have links to both the author's current CPAN author directory (their releases that are still on CPAN) and their BackPAN directory (all releases ever done).

I often want to look at both of these places, and always end up typing the URLs directly. This will save me some typing :-)


oalders:
Thanks for this @neilb.  There's currently a freeze on merging to master.  We'll get this in once the web site is using the Elasticsearch v2. :)

PR #1700, opened 2016-06-02
exclude t and xt directory from other files at search time

haarg:
We filter out any files in t or xt when separating out where to list
them on the release page.  If these files aren't filtered out in our
intial query, it can inflate the number of results over our 250 limit,
so that files we actually want to display won't be included.

This was the case for https://metacpan.org/release/OCBNET/CSS-Sass-3.3.4.
It has thousands of files in t/ named expected_output.css.  As a result,
most of the files we care about in the root weren't shown.


oalders:
Why did the push check fail while the PR check passed?

oalders:
👍

haarg:
I accidentally included an unrelated commit, then force pushed it away.  Travis tried to check that first push, but the commits weren't available by the time it started its run.

PR #1701, opened 2016-06-07
Link to reverse deps for module on module page

ilmari:
There already is a /requires/module/<module> endpoint, so use that on
individual module pages instead of /requires/release/<dist>.

In passing, clean up the variable passing to inc/dependencies.html by
passing the release object, since it uses several fields from it, not
just the dependencies.


oalders:
@ilmari would it be possible to make this a PR against `mi/es2`? We're trying to freeze master right now.

ilmari:
Recreated as #1721.

PR #1705, opened 2016-06-09
Remove erroneous trailing slashes from links

ilmari:
The canonical URLs for release and author info and source browsing don't
have trailing slashes, so avoid an unnecessary redirect roundtrip by
removing them from navigation links.


oalders:
@ilmari would it be possible to make this a PR against `mi/es2`? We're trying to freeze master right now.

ilmari:
Recreated against `mi/es2` as #1722.

Issue #1707, opened 2016-06-09
Author search auto-complete is under "Products" label


xsawyerx:
[This applies to v1 too.]

Select "Authors" in the search box radio button. Write some author's name so auto-complete kicks in. The names appear under the label "Products".

oalders:
@xsawyerx Thanks! Could you attach a screen shot?

xsawyerx:
![metacpan-product](https://cloud.githubusercontent.com/assets/54725/15933825/84493ca6-2e61-11e6-9fa1-55634c2d81d1.png)

oalders:
Wow.  I never noticed that before.  @danmcc is it possible to change that heading to "Authors"?

danmcc:
Definitely!  You can do it via CSS:

```
#autocomplete-direct-results:before {
  content: 'Authors'
}
```

Looks like `root/static/css/jquery.autocomplete.css` is the place to put it.

xsawyerx:
Fix merged. Closing.

Thank you. :)

PR #1714, opened 2016-06-09
Fix/autocomplete label

xsawyerx:
Per GH #1707, when searching for authors, the auto-complete shows the found results under a label "Products". This adds a CSS rule to fix it and show them under a label "Authors".


mickeyn:
Travis failures are not related to the change but to data issue we had in the API.
Verified manually - the 2 tests pass now.

oalders:
👍

PR #1722, opened 2016-06-10
Remove erroneous trailing slashes from links

ilmari:
The canonical URLs for release and author info and source browsing don't
have trailing slashes, so avoid an unnecessary redirect roundtrip by
removing them from navigation links.


oalders:
Thanks again. :)

PR #1723, opened 2016-06-10
[WIP] use config for links where possible

mickeyn:
probably doesn't cover all the issues we have, but will be easier to control from one config (for the templates) + some cleanup.

this is marked WIP - so feel free to comment/correct/patch

### DIFF
--- a/lib/MetaCPAN/Sitemap.pm
+++ b/lib/MetaCPAN/Sitemap.pm
@@ -58,7 +58,7 @@ sub process {
     # XXX Remove this hardcoded URL
     my $es = Search::Elasticsearch->new(
         cxn_pool         => 'Static::NoPing',
-        nodes            => ['api.metacpan.org'],
+        nodes            => ['api-v1.metacpan.org'],
         send_get_body_as => 'POST',
     );
 
--- a/root/about.html
+++ b/root/about.html
@@ -18,12 +18,12 @@ suggestions.
 
 MetaCPAN has two parts:
 
-* [metacpan.org](https://metacpan.org), a front end to...
-* [api-v1.metacpan.org](http://api-v1.metacpan.org/) the API
+* [<% mcpan_web %>](<% web_secure %>), a front end to...
+* [<% mcpan_api %>](<% api_secure %>) the API
 
 So
-[https://metacpan.org/module/Moose](https://metacpan.org/module/Moose)
-vs [http://api-v1.metacpan.org/module/Moose](http://api-v1.metacpan.org/module/Moose)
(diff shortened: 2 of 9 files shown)


lib/MetaCPAN/Sitemap.pm
```diff
@@ -58,7 +58,7 @@ sub process {
     # XXX Remove this hardcoded URL
     my $es = Search::Elasticsearch->new(
         cxn_pool         => 'Static::NoPing',
-        nodes            => ['api.metacpan.org'],
+        nodes            => ['api-v1.metacpan.org'],
```
oalders: Why not use the config here?

root/about.html
```diff
@@ -18,12 +18,12 @@
...
 
-* [metacpan.org](https://metacpan.org), a front end to...
-* [api-v1.metacpan.org](http://api-v1.metacpan.org/) the API
+* [<% mcpan_web %>](<% web_secure %>), a front end to...
+* [<% mcpan_api %>](<% api_secure %>) the API
 
 So
-[https://metacpan.org/module/Moose](https://metacpan.org/module/Moose)
-vs [http://api-v1.metacpan.org/module/Moose](http://api-v1.metacpan.org/module/Moose)
+[<% web_secure %>/module/Moose](<% web_secure %>/module/Moose)
```
oalders: I think we should use relative URLs wherever possible in the templates.  We could probably remove the variable from a lot of these.

lib/MetaCPAN/Sitemap.pm
mickeyn: no objection, will try to do that.
was just going for the easy ones first

mickeyn: how do I access/read the config from a non-controller module?

root/about.html
ranguard: The canonical urls in the templates SHOULD be hard coded - but thats just for SEO

lib/MetaCPAN/Sitemap.pm
oalders: This should work: https://github.com/CPAN-API/metacpan-web/blob/master/app.psgi#L58

root/about.html
oalders: How is this helpful for SEO?

ranguard: I only mean we don't want module/dist pages to have https://v1.metacpan.org/... as the canonical URL, or we loose google juice - in case google finds and starts to crawl v1

oalders: Can't we just block that entire subdomain via `robots.txt`?

PR #1724, opened 2016-06-11
Elasticsearch v0 to v2

oalders:
This is the branch we'll merge into master to close out the long lived `mi/es2` branch.


oalders:
@ranguard, @mickeyn  This is ready for merge now.

mickeyn:
that's great, shall we merge it to master then?

quick question: will it not affect the v0 servers in any way (in case of puppet running)?

ranguard:
I have switched bm-mc-01, bm-mc-02 and bm-mc-03 over to use: https://github.com/CPAN-API/metacpan-puppet/tree/legacy/v0 specifically note the `git_revision` https://github.com/CPAN-API/metacpan-puppet/blob/legacy/v0/hieradata/common.yaml#L88 https://github.com/CPAN-API/metacpan-puppet/blob/legacy/v0/hieradata/common.yaml#L103

I haven't actually run puppet under this new branch - but it should stop accidents

mickeyn:
alright then, here goes...

jberger:
\o/

PR #1737, opened 2016-06-22
fix more links

mickeyn:
fixed more hardcoded links to https://metacpan.org
relative where possible.

### DIFF
--- a/root/about.html
+++ b/root/about.html
@@ -18,12 +18,12 @@ suggestions.
 
 MetaCPAN has two parts:
 
-* [metacpan.org](https://metacpan.org), a front end to...
+* [v1.metacpan.org](https://v1.metacpan.org), a front end to...
 * [api-v1.metacpan.org](http://api-v1.metacpan.org/) the API
 
 So
-[https://metacpan.org/module/Moose](https://metacpan.org/module/Moose)
-vs [http://api-v1.metacpan.org/module/Moose](http://api-v1.metacpan.org/module/Moose)
+[https://v1.metacpan.org/module/Moose](https://v1.metacpan.org/module/Moose)
+vs [http://api.metacpan.org/module/Moose](http://api-v1.metacpan.org/module/Moose)
 
 MetaCPAN is [a community effort](/about/contributors), with all the code freely
 available on GitHub ([www](https://github.com/CPAN-API/metacpan-web),
--- a/root/static/opensearch.xml
+++ b/root/static/opensearch.xml
@@ -9,6 +9,6 @@
   <Language>en-us</Language>
   <SyndicationRight>open</SyndicationRight>
   <Image height="16" width="16" type="image/x-icon">https://metacpan.org/static/icons/favicon-16.ico</Image>
-  <Url type="text/html" template="https://metacpan.org/search?q={searchTerms}" />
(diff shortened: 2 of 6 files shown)


root/wrapper.html
```diff
@@ -39,14 +39,14 @@
...
         <%- IF rss %>
-        <link rel="alternate" type="application/rss+xml" title="RSS" href="https://metacpan.org/feed/<% rss %>" />
+        <link rel="alternate" type="application/rss+xml" title="RSS" href="/feed/<% rss %>" />
         <%- END %>
         <%- FOREACH css IN req.env.item('psgix.assets').grep(/\.css$/) %>
         <link href="<% css %>" rel="stylesheet" type="text/css">
         <%- END %>
         <link rel="search" href="/static/opensearch.xml" type="application/opensearchdescription+xml" title="MetaCPAN">
         <%- IF canonical %>
-        <link rel="canonical" href="https://metacpan.org<% canonical %>" />
```
ranguard: This is the one I don't think we should change - as v1.mc.org is NOT canonical, only mc.org is

root/about.html
```diff
@@ -18,12 +18,12 @@
 
 MetaCPAN has two parts:
 
-* [metacpan.org](https://metacpan.org), a front end to...
+* [v1.metacpan.org](https://v1.metacpan.org), a front end to...
```
ranguard: This doesn't need the v1 (we should be making this ready for production)

root/about.html
```diff
@@ -18,12 +18,12 @@
...
 * [api-v1.metacpan.org](http://api-v1.metacpan.org/) the API
 
 So
-[https://metacpan.org/module/Moose](https://metacpan.org/module/Moose)
-vs [http://api-v1.metacpan.org/module/Moose](http://api-v1.metacpan.org/module/Moose)
+[https://v1.metacpan.org/module/Moose](https://v1.metacpan.org/module/Moose)
```
ranguard: This shouldn't have the 'v1' here

root/about.html
```diff
@@ -18,12 +18,12 @@
...
+vs [http://api.metacpan.org/module/Moose](http://api-v1.metacpan.org/module/Moose)
```
ranguard: Change the link http://api.metacpan.org/module/Moose to http://api-v1.metacpan.org/module/Moose

root/static/opensearch.xml
```diff
@@ -9,6 +9,6 @@
   <Language>en-us</Language>
   <SyndicationRight>open</SyndicationRight>
   <Image height="16" width="16" type="image/x-icon">https://metacpan.org/static/icons/favicon-16.ico</Image>
-  <Url type="text/html" template="https://metacpan.org/search?q={searchTerms}" />
+  <Url type="text/html" template="/search?q={searchTerms}" />
```
ranguard: I think we might need this hard coded as it was - so that others using the opensearch know the URL to submit to?

mickeyn:
@ranguard care to patch it? :)

ranguard:
@mickeyn sure - I'll sort tonight

ranguard:
I think https://github.com/CPAN-API/metacpan-web/pull/1738 replaces this, but let me know if I'm wrong! - looks like several of the fixes were merged in but through another branch

PR #1738, opened 2016-06-23
Fix/links2 - make relative an explicit for new domain

ranguard:
should replace https://github.com/CPAN-API/metacpan-web/pull/1737


mickeyn:
great. will wait for travis results before merging.
thanks @ranguard

mickeyn:
@ranguard++

PR #1744, opened 2016-07-08
add missing abstracts in Release page

### LINKED ISSUE #1743: (v1) Abstracts not displayed on release pages
For example, compare https://v1.metacpan.org/release/JV/App-Music-ChordPro-0.63 to https://metacpan.org/release/JV/App-Music-ChordPro-0.63

mickeyn:
fixes #1743

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Release.pm
+++ b/lib/MetaCPAN/Web/Controller/Release.pm
@@ -122,14 +122,7 @@ sub view : Private {
 
     # Simplify the file data we pass to the template.
     my @view_files = map { single_valued_arrayref_to_scalar($_) }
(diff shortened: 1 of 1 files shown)


haarg:
Is there a reason we want to pick out individual fields rather than just everything that is in _source?

mickeyn:
TBH, I think you're right. I followed the example of the existing modules there.
The `_source` is a change from ES0.x to ES2.x where we have to pull `_source` separately and not through `fields` - so with the given query, it should be there.
I'll fix that.

mickeyn:
simplified :)
thanks @haarg

oalders:
Thanks!

PR #1746, opened 2016-07-11
use local links in pod

### LINKED ISSUE #1745: v1: Links in POD L<Some::Module> Points Wrongly
POD Links `L<Some::Module>` Points to https://metacpan.org...

For example in https://v1.metacpan.org/pod/AnyEvent:
    `An FAQ document is available as L<AnyEvent::FAQ>` -> "https://metacpan.org/pod/distribution/AnyEvent/lib/AnyEvent/FAQ.pod"

haarg:
Fixes #1745.

Requires metacpan/metacpan-api#506.


oalders:
👍

Issue #1692, opened 2016-05-14
The "Browse CPAN directory" link on author pages is really a BackPAN link


neilb:
When you look at an author page, at the bottom left there's a link "Browse CPAN directory". Eg when looking at [the author page for AVERNA](https://metacpan.org/author/AVERNA), you get a link to:

> [cpan.metacpan.org/authors/id/A/AV/AVERNA/](https://cpan.metacpan.org/authors/id/A/AV/AVERNA/)

But that's not a CPAN author directory, it's a BackPAN author directory. It would be neat to have both:

 * Browse CPAN directory
 * Browser BackPAN directory

So you can see everything they've released, and everything they've currently got on CPAN. The CPAN link could just link to [cpan.org/authors/id/A/AV/AVERNA](https://cpan.org/authors/id/A/AV/AVERNA)

I'm guessing / hoping this is a quick change, so will have a look right after I submit this.

neilb:
Here's my pull request to make the CPAN directory link stop lying, and to add a BackPAN link:

> [pull/1693](https://github.com/CPAN-API/metacpan-web/pull/1693)

book:
In the case an author has deleted all of their releases (e.g. [JFENAL](https://metacpan.org/author/JFENAL), it feels like the BackPAN link should be more prominent, or that maybe the author page should have a message stating their releases have been deleted and are available on BackPAN.

neilb:
> maybe the author page should have a message stating their releases have been deleted and are available on BackPAN

That would be a nice touch. But right now that  would need an additional lookup. If you look at the author information:

> https://fastapi.metacpan.org/v1/author/BOOK

There's nothing in there about how much stuff you've got on CPAN or BackPAN. It'd be handy to have that information in here:

 * number of distributions and/or releases on CPAN
 * number of distributions and/or releases on BackPAN

With that info in the author return, it wouldn't need an additional query to implement @book's suggestion.

oalders:
That wouldn't be very hard to do.

neilb:
@oalders do you want me to open a separate issue for that (and then you could put volunteer needed on that)?

oalders:
@neilb Please.  That would be less confusing. :)  I'll close this issue now then.

PR #1750, opened 2016-07-15
author page: cpan/backpan counts + backpan-only message

mickeyn:
per https://github.com/metacpan/metacpan-api/issues/502 + #1692


oalders:
Thanks!

PR #1751, opened 2016-07-18
move code from template to the controller module

mickeyn:
moved some structure building from the template (why? why? why?) to the module.

pretty much translated as is, so **feel free to comment or fix**.

for some reason, it also seems to solve this:
03:12:53 � Grinnz� something funky here https://v1.metacpan.org/release/DVINCI/Mojolicious-Command-generate-controller-0.01
03:13:01 � Grinnz� looks normal under v0, just one provides lol

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Release.pm
+++ b/lib/MetaCPAN/Web/Controller/Release.pm
@@ -9,6 +9,7 @@ use namespace::autoclean;
 
 BEGIN { extends 'MetaCPAN::Web::Controller' }
 use List::Util ();
+use Ref::Util qw( is_arrayref );
 
 with qw(
     MetaCPAN::Web::Role::ReleaseInfo
@@ -125,6 +126,8 @@ sub view : Private {
         map +{ %{ $_->{fields} }, %{ $_->{_source} }, },
         @{ $modules->{hits}->{hits} };
 
+    my $categories = $self->_files_to_categories( $out, \@view_files );
+
     my $changes
         = $c->model('API::Changes')->last_version( $reqs->{changes}, $out );
 
@@ -139,6 +142,11 @@ sub view : Private {
         root     => \@root_files,
         examples => \@examples,
         files    => \@view_files,
+
+        documentation     => $categories->{documentation},
+        documentation_raw => $categories->{documentation_raw},
+        provides          => $categories->{provides},
+        modules           => $categories->{modules},
 
         # TODO: Put this in a more general place.
         # Maybe make a hash for feature flags?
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Web/Controller/Release.pm
```diff
@@ -158,6 +165,56 @@ sub view : Private {
     );
 }
 
+sub _files_to_categories {
```
oalders: For the sake of consistency, it would be nice not to mix functions and methods in the same class.  So, we could either call this via `$self->` or we could move it into some kind of a helper class that exports this function.

lib/MetaCPAN/Web/Controller/Release.pm
```diff
@@ -158,6 +165,56 @@ sub view : Private {
...
+    my $ret = +{
+        provides          => [],
+        documentation     => [],
+        documentation_raw => [],
+        modules           => [],
+    };
+
+    for my $f ( @{$files} ) {
+        if ( length $f->{module} ) {
+            push @{ $ret->{modules} } => $f if $f->{documentation};
```
oalders: Any particular reason for the fat comma here?  I think a plain old comma might be easier to read.

lib/MetaCPAN/Web/Controller/Release.pm
```diff
@@ -158,6 +165,56 @@ sub view : Private {
...
+            my %info = (
+                status  => $f->{status},
+                path    => $f->{path},
+                release => $release->{name},
+                author  => $release->{author},
+            );
+            for my $m (
+                ref $f->{module} eq 'ARRAY'
```
oalders: You could use `Ref::Util` here.  ;)

oalders:
Thanks for moving the logic out of the template.  There was way too much going on there!

lib/MetaCPAN/Web/Controller/Release.pm
mickeyn: will fix

lib/MetaCPAN/Web/Controller/Release.pm
mickeyn: just my style (I like it with push, join) - i'll change that

lib/MetaCPAN/Web/Controller/Release.pm
mickeyn: right :)

mickeyn:
i'll look for other candidates as well

mickeyn:
I updated the code according to all your comments, thanks for the tips

oalders:
Thanks.  Is this still a WIP?

mickeyn:
not if you're OK with it :) i'll update the title

haarg:
Would be good to squash the commits.

mickeyn:
done

oalders:
I guess we should enable squashing: https://github.com/blog/2141-squash-your-commits

oalders:
Thanks!

PR #1754, opened 2016-07-25
add datacenters for mirrors.cpan.org to use

ranguard:
This will allow mirrors.cpan.org to get an up-to-date list of Fastly servers so that they can all be geo-located and spread around the world.

Also added to the CPAN list, which cpan.metacpan.org isn't at the moment!

### DIFF
--- a/lib/MetaCPAN/Web/Role/Fastly.pm
+++ b/lib/MetaCPAN/Web/Role/Fastly.pm
@@ -86,6 +86,17 @@ has 'cdn_times' => (
     lazy_build => 1,
 );
 
+sub datacenters {
+    my ($c) = @_;
+    my $net_fastly = $c->_net_fastly();
+    return unless $net_fastly;
+
+    # Uses the private interface as fastly client doesn't
+    # have this end point yet
+    my $datacenters = $net_fastly->client->_get('/datacenters');
+    return $datacenters;
+}
+
 sub _build_cdn_times {
     return {
         one_hour => 3600,
--- a/lib/MetaCPAN/Web/Controller/SysAdmin/DataCenters.pm
+++ b/lib/MetaCPAN/Web/Controller/SysAdmin/DataCenters.pm
@@ -1,0 +1,27 @@
+package MetaCPAN::Web::Controller::SysAdmin::DataCenters;
+
+use MetaCPAN::Moose;
+
+BEGIN { extends 'MetaCPAN::Web::Controller' }
+
+# This is used by http://mirrors.cpan.org/ fetch our
+# datacenters and update the official CPAN list
+
+sub list_datacenters : Path('list') : Args(0) GET {
+    my ( $self, $c ) = @_;
+
+    # Fetch the data center info from fastly
+    # XXX: will not work unless you have the credentials
+    my $datacenters = $c->datacenters;
+
+    $c->add_surrogate_key('datacenters');
+    $c->cdn_cache_ttl( $c->cdn_times->{one_day} );
+    $c->res->header(
+        'Cache-Control' => 'max-age=' . $c->cdn_times->{one_day} );
+    $c->stash( { success => $datacenters } );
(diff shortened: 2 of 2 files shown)


lib/MetaCPAN/Web/Controller/Sysadmin/Datacenters.pm
```diff
@@ -0,0 +1,29 @@
+package MetaCPAN::Web::Controller::Sysadmin::Datacenters;
```
oalders: Could we s/Sysadmin::Datacenters/SysAdmin::DataCenters/?  I know it's picky, but if we do this later a lot of people with checkouts on OS X are going to be irritated by the change in case. ;)

lib/MetaCPAN/Web/Controller/Sysadmin/Datacenters.pm
```diff
@@ -0,0 +1,29 @@
...
+
+use Moose;
```
oalders: Have you tried `use MetaCPAN::Moose` here?  Am curious if it causes any problems.

lib/MetaCPAN/Web/Controller/Sysadmin/Datacenters.pm
```diff
@@ -0,0 +1,29 @@
...
+    my $datacenters = $c->datacenters;
+
+    $c->add_surrogate_key('datacenters');
+    $c->cdn_cache_ttl( $c->cdn_times->{one_day} );
+    $c->res->header(
+        'Cache-Control' => 'max-age=' . $c->cdn_times->{one_day} );
+
+    $c->stash( { success => $datacenters } );
+    $c->detach( $c->view('JSON') );
+
```
oalders: Could clean up this extra newline.

lib/MetaCPAN/Web/Role/Fastly.pm
```diff
@@ -86,6 +86,15 @@ has 'cdn_times' => (
     lazy_build => 1,
 );
 
+sub datacenters {
+    my ($c) = @_;
+    my $net_fastly = $c->_net_fastly();
+    return unless $net_fastly;
+
+    my $datacenters = $net_fastly->client->_get('/datacenters');
```
oalders: Why are we calling a private method here?

lib/MetaCPAN/Web/Controller/Sysadmin/Datacenters.pm
ranguard: ok

lib/MetaCPAN/Web/Controller/Sysadmin/Datacenters.pm
ranguard: Ahh, no I started this in mc-api and that uses `Moose` not `MetaCPAN::Moose` - I didn't notice when I moved it over (mc-api is to ES specific!)

lib/MetaCPAN/Web/Controller/Sysadmin/Datacenters.pm
ranguard: heh - dude you _are_ picky! - but sure :)

lib/MetaCPAN/Web/Role/Fastly.pm
ranguard: Because they don't have a public one ;) - I'll comment

lib/MetaCPAN/Web/Controller/Sysadmin/Datacenters.pm
ranguard: MC::Moose sucks... you have to `use strict` and `use warnings` for perl critic tests

oalders: It's actually our critic config that sucked.  I've now made it suck less via ae6573d6973

PR #1765, opened 2016-08-28
Leo/fastly add keys for caching

ranguard:
Do not merge - yet - API needs to be updated to purge correct fastly service


haarg:
I already mentioned this in IRC, but I'll put it here so it's in a more permanent form.  Some changes I'd like to see: 
- dist and author should be namespaced, such as "DIST-$dist" and "AUTHOR-$author"
- a common surrogate key for all html pages
- a common surrogate key for all static files (images etc)
- a common surrogate key for all "asset" files (combined/minified css and js)

When we change the CSS or JS, we would purge the html pages.  The assets should never need to be purged.  The static files could be purged as needed (although with the browser TTL on these, it is unlikely to be effective).

cpanfile
```diff
@@ -57,13 +58,15 @@ requires 'Moo', '2.000002';
 requires 'MooseX::ClassAttribute';
 requires 'MooseX::Role::Parameterized', '1.02';
 requires 'MooseX::StrictConstructor';
+requires 'MooseX::Fastly::Role';
```
oalders: Maybe we could make it a habit to include minimum version numbers when adding to the `cpanfile`?

lib/MetaCPAN/Web/Controller/Account/Favorite.pm
```diff
@@ -26,7 +26,12 @@ sub add : Local : Args(0) {
         $res = $model->add_favorite( $data, $c->token )->recv;
     }
 
-    # Clear Fastly.. something changed
+    # TODO: validate these values?
```
oalders: How did you want to validate these?

lib/MetaCPAN/Web/Controller/Feed.pm
```diff
@@ -13,9 +13,17 @@ use Text::Markdown qw/markdown/;
 use Importer 'MetaCPAN::Web::Elasticsearch::Adapter' =>
     qw/ single_valued_arrayref_to_scalar /;
 
-sub recent : Local : Args(0) {
+sub feed_index : PathPart('feed') : Chained('/') : CaptureArgs(0) {
     my ( $self, $c ) = @_;
+
```
oalders: Extra newline.

lib/MetaCPAN/Web/Controller/Recent.pm
```diff
@@ -17,7 +17,11 @@ sub index : Path : Args(0) {
         ->recent( $req->page, $page_size, $req->params->{f} || 'l' )->recv;
     my $latest = [ map { $_->{fields} } @{ $data->{hits}->{hits} } ];
     single_valued_arrayref_to_scalar($latest);
-    $c->res->last_modified( $latest->[0]->{date} ) if (@$latest);
+
+    $c->add_surrogate_key('RECENT');
+    $c->browser_max_age('1m');
+    $c->cdn_max_age('1m');    # tough to paging for now
```
oalders: Not sure what "tough to paging" means?

t/controller/pod.t
```diff
@@ -25,6 +25,11 @@ test_psgi app, sub {
 
     my $latest = $res->content;
     ok( $res = $cb->( GET $this ), "GET $this" );
+    is(
+        $res->headers->header('Surrogate-Key'),
+        'dist=MOOSE author=ETHER content_type=text/html content_type=text',
+        'Surrogate-Key dist/author/content typ'
```
oalders: Should this be "type" rather than "typ"?

oalders:
Looks great.  Thanks for including the tests.  @haarg, did you want to look this over?

lib/MetaCPAN/Web/Controller/Account/Favorite.pm
ranguard: Hang over from development, they are using a standard method under the hood now, so they will be the same format - have removed the comment, thanks

lib/MetaCPAN/Web/Controller/Recent.pm
ranguard: You _might_ get something appearing at the end of page 1, and the start of page 2, if you happen to hit them at different - BRAIN WAVE! - api now purges 'DIST_UPDATES' any time there is an update to any dist, so we can cache this for a year - pull request will reflect this shortly

ranguard:
@haarg please also look at https://github.com/metacpan/p5-metacpan-role/blob/master/lib/MetaCPAN/Role/Fastly/Catalyst.pm#L57 that won't work for anything outside of Catalyst, but will work for anything Catalyst serves.

Issue #1767, opened 2016-08-31
V1: duplicate entries in release page


mickeyn:
@shadowcat-mst reported this on IRC:

17:02:03 � mst� https://v1.metacpan.org/release/REEDFISH/Net-FullAuto-1.0000233
17:02:11 � mst� why is Rem_Command showing repeatedly for the same file?

mickeyn:
fixed

PR #1768, opened 2016-08-31
controller/release: correct categories logic

mickeyn:
Code originally migrated from template. attempt to cut a corner was buggy.
This makes it a bit more accurate, but this code still need some love.

At least now the page will look like it does in v0.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Release.pm
+++ b/lib/MetaCPAN/Web/Controller/Release.pm
@@ -176,34 +176,41 @@ sub _files_to_categories {
     };
 
     for my $f ( @{$files} ) {
-        if ( length $f->{module} ) {
-            push @{ $ret->{modules} }, $f if $f->{documentation};
-            my %info = (
-                status  => $f->{status},
-                path    => $f->{path},
-                release => $release->{name},
-                author  => $release->{author},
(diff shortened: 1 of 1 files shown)


mickeyn:
this will solve #1767

lib/MetaCPAN/Web/Controller/Release.pm
```diff
@@ -176,35 +176,42 @@ sub _files_to_categories {
...
-                    push @{ $ret->{provides} }, \%info;
-                }
+        my %info = (
+            status  => $f->{status},
+            path    => $f->{path},
+            release => $release->{name},
+            author  => $release->{author},
+        );
+
+        my @modules
```
ranguard: As this is the edge case, might be nice to add comment about the situation in which it happens, even add an e.g.?

mickeyn: as this is the first step of migrating code out of templates, i'd still wish to rework it as i don't find it good enough... just a 1-1 translation.

ranguard: For me ANY comments would be helpful - but JFDI is good as well. :)

PR #1776, opened 2016-09-27
author page: moved logic from template to module

mickeyn:
another structure being built in the templates moved to the controller.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Author.pm
+++ b/lib/MetaCPAN/Web/Controller/Author.pm
@@ -73,15 +73,18 @@ sub index : Chained('root') PathPart('') Args(0) {
         @$releases;
     $c->res->last_modified($date) if $date;
 
+    my ( $aggregated, $latest ) = @{ $self->_calc_aggregated($releases) };
+
     $c->stash(
         {
-            author      => $author,
-            faves       => $faves,
-            releases    => $releases,
-            show_author => 1,
-            template    => 'author.html',
-            took        => $took,
-            total       => $data->{hits}->{total},
+            aggregated => $aggregated,
+            author     => $author,
+            faves      => $faves,
+            latest     => $latest,
+            releases   => $releases,
+            template   => 'author.html',
+            took       => $took,
+            total      => $data->{hits}->{total},
         }
     );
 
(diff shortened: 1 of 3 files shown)


lib/MetaCPAN/Web/Controller/Author.pm
```diff
@@ -73,15 +73,18 @@ sub index : Chained('root') PathPart('') Args(0) {
...
-            took        => $took,
-            total       => $data->{hits}->{total},
+            author     => $author,
+            faves      => $faves,
+            releases   => $releases,
+            template   => 'author.html',
+            took       => $took,
+            total      => $data->{hits}->{total},
+            aggregated => $aggregated,
+            latest     => $latest,
```
oalders: Could we alpha sort the keys?

lib/MetaCPAN/Web/Controller/Author.pm
```diff
@@ -129,6 +132,29 @@ sub releases : Chained('root') PathPart Args(0) {
...
+        } ( $rel->{date}, $latest->{date} );
+        $latest = $rel if $canon_rel > $canon_lat;
+
+        next if $last and $last eq $rel->{distribution};
+        $last = $rel->{distribution};
+        next unless $rel->{name};
+        push @aggregated, $rel;
+    }
+
+    return +[ \@aggregated, $latest ];
```
oalders: What's the reason for the `+` here?

mickeyn: ah, just style (got used to write +[] and +{} to indicate refs), nothing except readability

lib/MetaCPAN/Web/Controller/Author.pm
mickeyn: sure

mickeyn:
@oalders updated code according to your comments

lib/MetaCPAN/Web/Controller/Author.pm
oalders: I can see it for 
```perl
sub { +{ foo => 'bar' } }
```

since that makes the braces easier to read, but with the array reference it mostly confuses me.  :)

PR #1778, opened 2016-09-28
[WIP] use /author/by_user endpoint instead of ES query

mickeyn:
this PR is to make use of https://github.com/metacpan/metacpan-api/pull/530 on the web side (-1 ES query)

### DIFF
--- a/lib/MetaCPAN/Web/Model/API/Author.pm
+++ b/lib/MetaCPAN/Web/Model/API/Author.pm
@@ -37,7 +37,7 @@ sub search {
     my ( $self, $query, $from ) = @_;
     my $cv = $self->cv;
     $from ||= 0;
-    $self->request("/author/search?key=$query&from=$from&size=10")->cb(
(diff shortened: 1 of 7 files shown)


lib/MetaCPAN/Web/Model/API/Author.pm
```diff
@@ -30,81 +30,27 @@ sub get {
...
-            query => {
-                constant_score => {
-                    filter => { ids => { values => [ map {uc} @author ] } }
-                }
-            },
-            size => scalar @author,
-        }
-    );
-
+    return $self->request( '/author/by_id?id=' . join ',', @author );
```
haarg: These should be passed as multiple id values, not a comma separated value.

Also, `->request` accepts a hashref of query parameters to add.  If that isn't adequate, we should extend it so this can be passed in in a structured manor.  The other changes in this PR have the same issue.

mickeyn:
most of this is currently happening in small chunks, closing the PR

Issue #1766, opened 2016-08-29
HTML-Table page seems to be permanently borked


andk:
https://metacpan.org/release/HTML-Table gives me a "please come back later" page. I got the link to it from metacpan itself: https://metacpan.org/author/AJPEACOCK. I get the "please come back later page already for more than a day.

As far as I can see, HTML::Table is AJPEACOCK/HTML-Table-2.08a.tar.gz which contains version "2.08a" but is indexed by PAUSE as having "2.08". This may be completely irrelevant for the issue, I just want to mention it.

oalders:
@andk thanks for reporting. Looks like it's broken on v1 as well.

ranguard:
Now working on v1 - thanks @haarg

PR #1780, opened 2016-10-01
fix latest changes when dist version is unparsable

haarg:
If the dist version is not parsable as a version, just do string
comparisons against the versions found in the change log.

Fixes 500 errors on https://metacpan.org/release/WHITEWIND/neverbounce_0_02


haarg:
Also fixes #1766

PR #1791, opened 2016-11-17
change signature on request to allow query params

haarg:
API->request accepted a hashref of parameters, and in some places they
were used as if they were URL params.  Instead, they only accepted a
method and an access token to use.

Change request to use the parameters as URL query parameters, and use a
separate parameter for the URL.


oalders:
Thanks!

PR #1792, opened 2016-11-18
Short-circuit correctly in search_expanded

tsibley:
The total will always be the number of documents matched, not the number
of documents returned for the given page.  Instead, do what
search_collapsed does and check if we have any distributions.

Additionally, guard search_descriptions against an empty id list, which
produces invalid queries.

I found these while tracking down the cause of thousands of 500s on
/file/_search, ultimately being triggered by a web crawler.


tsibley:
This is now passing that I've tidied.

PR #1795, opened 2016-11-18
cpanfile.snapshot builds

zdykstra:
Attempt to build on 5.22 with cpanfile.snapshot and with out it. This is skipped for 5.18 and 5.20, where it only builds with a base cpanfile


oalders:
Thanks!

PR #1809, opened 2016-11-19
Feed URLs are no longer relative

oalders:


tsibley:
Hmm.  The two commits seem like they should be one, as the second one makes a bunch of changes and fixes to the code touched by the first commit.

This also only fixes `/feed/recent` for me. `/feed/author/OALDERS` returns releases and favorites with relative URLs and so does `/feed/news`.

tsibley:
:+1: :ship:

PR #1822, opened 2016-11-24
Web search uses api endpoint

jberger:
NOTE! I want people to pay careful attention to this change, try it, try to break it. Search is one of the most important parts of the site and as such I'm cautious.

Hopefully, if I've done my job correctly the results should be exactly the same as currently. I have not exposed the boosts etc.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Search.pm
+++ b/lib/MetaCPAN/Web/Controller/Search.pm
@@ -66,21 +66,7 @@ sub index : Path : Args(0) {
         }
     }
     else {
-        my $user = $c->user_exists ? $c->user->id : undef;
-
-        # these would be nicer if we had variable-length lookbehinds...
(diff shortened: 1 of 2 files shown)


jberger:
Also, pay attention to the slight change in `single_dist` behavior (which defines if the "search in distribution" link is shown on the results). Before it checked to see if the (munged) query contained a `distribution:` limiter. This meant that the search was already within a limitation to just a distribution. As the web side no longer has access to the munged query some change had to be made.

The api currently (if not explicitly told) decides whether it should show expanded to collapsed results based on [this regex](https://github.com/metacpan/metacpan-api/blob/d174bff46f17d8c537ae0cee33e8d64d0d722c0b/lib/MetaCPAN/Model/Search.pm#L59) which is similar. It then contains within the result a key saying whether the results were collapsed. It is this value that is then used to decide if the "search in distribution" link is shown. Is this sufficient or does it need tweaking?

PR #1836, opened 2016-12-20
local author autocomplete

### LINKED ISSUE #1550: Large amount of whitespace at the top
There's a rather large amount of whitespace at the top of the page.

I hadn't noticed this until I was looking at #1548.
The whitespace seems to be present without that addition, but the extra row of input fields makes the area look even larger.

screenshot - 06202015 - 08 34 35 am

Is it just my eyes, or does anyone else agree that that's too much space?

haarg:
This implements author autocomplete locally, removing constructor.io and the radio buttons.

In the future, we could switch to constructor.io for the entire thing, but it requires more work.  I don't think the current author autocomplete is used enough to justify keeping it, even if constructor.io could provide better results.

This also removes the "15 results (0.013 seconds)" text from pages without pagination, and moves it to below the pagination controls.  I don't think it provides useful information on pages like the release page, and without it the layout is more consistent.

The spacing at the top of the page is also reduced slightly.

Fixes #1550, and addresses the same issues as #1569.
Screenshots:

Module page before:
screen shot 2016-12-20 at 4 57 01 pm
After:
screen shot 2016-12-20 at 4 58 24 pm

Release page before:
screen shot 2016-12-20 at 4 57 26 pm
After:
screen shot 2016-12-20 at 4 58 41 pm

Autocomplete before:
screen shot 2016-12-20 at 4 57 48 pm
After:
screen shot 2016-12-20 at 4 59 03 pm


haarg:
This also makes changes to the web's autocomplete end point.  If anyone is using that externally, it will break their code.  The api provides its own autocomplete end point though, so anyone trying to do autocomplete should be using that instead.

oalders:
Excellent!

Issue #1840, opened 2017-01-05
th tags and style attributes stripped from HTML


benkasminbullock:
Within =begin HTML blocks in POD, <th> tags and style attributes are being removed from the html before it's displayed.

Style tags removed here:

https://metacpan.org/source/BKB/HTML-Make-0.07/lib/HTML/Make.pod#L237

TH tags removed here:

https://metacpan.org/source/BKB/HTML-Make-0.07/lib/HTML/Make.pod#L57
https://metacpan.org/source/BKB/HTML-Make-0.07/lib/HTML/Make.pod#L410

See the output here:

view-source:https://metacpan.org/pod/release/BKB/HTML-Make-0.07/lib/HTML/Make.pm

For example, Item Cost ends up like this:

    <table>
    <tr>ItemCost<tr><td>camera</td>

oalders:
Sounds like we need to tweak the `HTML::Restrict` args.

haarg:
Allowing th tags should be fine, but style attributes are stripped intentionally.

benkasminbullock:
> Allowing th tags should be fine, but style attributes are stripped intentionally.

The relevant code appears to be this:

https://github.com/metacpan/metacpan-web/blob/dbd014b4fdd014860d35a9f95a1365deb30c4219/lib/MetaCPAN/Web/Controller/Pod.pm#L164

It appears `th` is omitted and yet the `style` attribute appears to be permitted on `td` and `tr` and `table`, but is not allowed on the `li` element which was used in my html block:

https://github.com/metacpan/metacpan-web/blob/dbd014b4fdd014860d35a9f95a1365deb30c4219/lib/MetaCPAN/Web/Controller/Pod.pm#L147

https://metacpan.org/source/BKB/HTML-Make-0.07/lib/HTML/Make.pod#L245

I can't see what logical argument there is for stripping style attributes from li while not doing the same for td and tr and table.

haarg:
I'd argue that we should be stripping it (or heavily restricting it) from the existing places it is allowed.

benkasminbullock:
Regarding this, due to the issues both with metacpan.org and with search.cpan.org, I decided to omit HTML blocks from the module. See the following for the problem with search.cpan.org:

http://search.cpan.org/~bkb/HTML-Make-0.07/lib/HTML/Make.pod

The HTML blocks also do not display with Pod::POM::Web.

So as far as the problems with my module are concerned it's OK to close this issue.

[...]

PR #1844, opened 2017-01-11
pod filtering

haarg:
This makes our pod filtering a bit more strict, not allowing style tags and allowing less id attributes.  It also allows the th tag, and adjusts img tags to allow paths relative to the pod file being displayed.

Classes on divs are updated to correspond to metacpan/metacpan-api#611


haarg:
Addresses #1840

oalders:
👍

Issue #1801, opened 2016-11-19
Dashboard not working


ranguard:
https://metacpan.org/lab/dashboard 

hi @szabgab - we are now on the new API, and this doesn't seem to be working, if you get time to have a look and fix that would be great - thanks.

barefootcoder:
I assume I'm seeing the same thing as @ranguard ... specifically, when I go to the dashboard page, I get this message:
```
The dashboard is only available to people who have logged in to MetaCPAN. 
```
It _used_ to work, although I'll admit I can't remember how long ago that was.  Not a big deal, but I remember the dashboard being useful, so it might be nice to have back.

haarg:
There are two issues at the moment.  First is that it isn't recognizing when you are logged in.  That should be trivial to fix.  Once we fix that though, there is something else broken that I haven't looked at.

ranguard:
hi @szabgab are you interested in fixing this? - otherwise I'll hide it until someone does. @barefootcoder  & @haarg - it was always experimental, probably the query got broken with the upgrade to ES 2.4 or something

mickeyn:
I can try to fix the query to work on ES 2.4

PR #1850, opened 2017-02-12
lab/dashboard: update ES query to work on v1

mickeyn:
not properly tested, but the query seems to work now.
it's broken anyway so I don't see any risk in merging - let's just wait till tests pass.


mickeyn:
this should solve #1801

oalders:
👍

PR #1851, opened 2017-02-13
don't un-json `extra` if it's the empty string

simbabque:
The default value for the extra field is the empty string, because it's an empty text box. But it's supposed to be JSON and it was always decoded from JSON, which produced an exception.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Account.pm
+++ b/lib/MetaCPAN/Web/Controller/Account.pm
@@ -52,21 +52,24 @@ sub profile : Local : Args(0) {
     return unless ( $req->method eq 'POST' );
 
     my $data = $author;
-    $data->{blog} = $req->param('blog.url')
+    $data->{blog}
+        = $req->param('blog.url')
         ? [
         pairwise { { url => $a, feed => $b } }
         @{ [ $req->param('blog.url') ] },
         @{ [ $req->param('blog.feed') ] }
         ]
         : undef;
-    $data->{donation} = $req->param('donation.name')
+    $data->{donation}
+        = $req->param('donation.name')
         ? [
         pairwise { { name => $a, id => $b } }
         @{ [ $req->param('donation.name') ] },
         @{ [ $req->param('donation.id') ] }
         ]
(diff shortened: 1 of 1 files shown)


mickeyn:
tidy test fails over some weird edge-case we struggled with before.
please revert the tidy changes and leave only yours, otherwise it will keep failing on travis.

lib/MetaCPAN/Web/Controller/Account.pm
```diff
@@ -52,21 +52,24 @@ sub profile : Local : Args(0) {
     return unless ( $req->method eq 'POST' );
 
     my $data = $author;
-    $data->{blog} = $req->param('blog.url')
+    $data->{blog}
```
mickeyn: this is a weird tidyall issue (probably version mismatch), this change fails on Travis.

lib/MetaCPAN/Web/Controller/Account.pm
```diff
@@ -52,21 +52,24 @@ sub profile : Local : Args(0) {
...
+        = $req->param('blog.url')
         ? [
         pairwise { { url => $a, feed => $b } }
         @{ [ $req->param('blog.url') ] },
         @{ [ $req->param('blog.feed') ] }
         ]
         : undef;
-    $data->{donation} = $req->param('donation.name')
+    $data->{donation}
```
mickeyn: same

lib/MetaCPAN/Web/Controller/Account.pm
```diff
@@ -52,21 +52,24 @@ sub profile : Local : Args(0) {
...
+        = $req->param('donation.name')
         ? [
         pairwise { { name => $a, id => $b } }
         @{ [ $req->param('donation.name') ] },
         @{ [ $req->param('donation.id') ] }
         ]
         : undef;
-    $data->{profile} = $req->param('profile.name')
+    $data->{profile}
```
mickeyn: same

mickeyn:
thanks

PR #1852, opened 2017-02-18
Unittest/account controller

simbabque:
This should cover everything but the identities action. I didn't do that one because I've not yet interacted with it. I also fixed a bug I introduced with my previous PR when I fixed that other bug... :)


simbabque:
The coverage has gone down because there are now more files being covered than before, but not at 100% percent. At least that's what should be the reason, but the coverage report is strange and seems to not include all files anyway.

oalders:
Thanks @simbabque!

PR #1857, opened 2017-02-26
include metacpan user id in account profile

### LINKED ISSUE #1199: Allow users to view their user_id
Please let the site's users view their user id, so that they will then be able to ask 3rd-party sites & services to fetch their MetaCPAN favorites or interact in other ways.

One such service is https://www.perlmodules.net/, where retrieving one's own list of favorites would be quite useful.

So, exposing the user id under https://metacpan.org/account/profile might be the most appropriate place.

simbabque:
Since metacpan is in itself also a profile that's identified by a user id I figured it makes sense to put this under _Profiles_. It could either go as the first or the last one. Since new additions always go to the bottom I figured it makes more sense to show it at the top.

I decided to create a copy of the metacpan icon in root/static/images/profile to keep consistent with all the other profile images.

Not sure if perlmodules.net is actually using the user id already, but if they are (and possibly others also do) it might make sense to include an FAQ entry that explains what the user id can be used for and where to find it.

Closes #1199


oalders:
Thanks @simbabque!

PR #1860, opened 2017-03-19
fix login for non-pause users

### LINKED ISSUE #1819: authentication via google/twitter/github is broken
Hi I am trying to authenticate ( get logged in to my profile ). to metacpan.org

1. when using authentication via my github account the system silently return me back to the current page and I still see at the up right corner that I am not logged in ( sign i link  still appears )

2. the same behavior for attempt with twitter account

3. when using authentication vua google account a I redirected to [page](https://accounts.google.com/o/oauth2/auth?client_id=265904441292.apps.googleusercontent.com&response_type=code&redirect_uri=https%3A%2F%2Ffastapi.metacpan.org%2Flogin%2Fgoogle&scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fuserinfo.profile) with content:

```
400. That’s an error.

Error: redirect_uri_mismatch

The redirect URI in the request, https://fastapi.metacpan.org/login/google, does not match the ones authorized for the OAuth client. Visit https://console.developers.google.com/apis/credentials/oauthclient/265904441292.apps.googleusercontent.com?project=265904441292 to update the authorized redirect URIs.

Learn more
Request Details

    scope=https://www.googleapis.com/auth/userinfo.profile
    response_type=code
    redirect_uri=https://fastapi.metacpan.org/login/google
    client_id=265904441292.apps.googleusercontent.com

That’s all we know.
```

haarg:
Logging in as a user that didn't have a PAUSE account connected was
working, but the code to fetch the favorites list (which controls the
login menu) wasn't interpreting that correctly.  The favorites code was
checking for a "profile", which only exists for PAUSE users.

Instead, use the normal authentication routines and the resulting user
object.

Fixes #1819

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Account/Favorite.pm
+++ b/lib/MetaCPAN/Web/Controller/Account/Favorite.pm
@@ -83,8 +83,7 @@ sub _add_fav_list_to_stash {
 
     my $user = $c->user;
 
-    my $faves_cv
-        = $c->model('API::Favorite')->by_user( $user->id, $size );
+    my $faves_cv   = $c->model('API::Favorite')->by_user( $user->id, $size );
     my $faves_data = $faves_cv->recv;
     my $faves      = [
         sort { $b->{date} cmp $a->{date} }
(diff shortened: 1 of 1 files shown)


sergeykolychev:
@haarg I believe  do have pause account associated  with my profile, so this is probably not a root of the problem.

sergeykolychev:
@haarg The fix is definitely correct one. Though I had my pause account associated with my metacpan account something was wrong in this association, unlinking my pause account and linking it again seem to have fixed the login to the site for me.

oalders:
@sergeykolychev and @haarg thanks very much to both of you for getting this sorted!

PR #1872, opened 2017-04-26
releaseinfo bug fix

mickeyn:
this fixes current failing tests on master and this issue:
21:06:41 � Grinnz� https://metacpan.org/pod/indirect throwing an error


oalders:
LGTM

PR #1874, opened 2017-05-05
add class "img-responsive" to <img> tag in pod

skaji:
Now style attributes for `` tags are not allowed (cf #1844).

Then `` can be bigger than the parent element.
See, for example, https://metacpan.org/pod/release/SKAJI/Parallel-Pipes-0.002/lib/Parallel/Pipes.pm

This pull request adds class "img-responsive" to `` tag in pod so that images are made responsive-friendly.
See also http://getbootstrap.com/css/#images

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Pod.pm
+++ b/lib/MetaCPAN/Web/Controller/Pod.pm
@@ -168,7 +168,7 @@ sub view : Private {
         },
         replace_img => sub {
             my ( $tagname, $attrs, $text ) = @_;
-            my $tag = '<img';
+            my $tag = '<img class="img-responsive"';
             for my $attr (qw( alt border height width src title)) {
                 next
                     unless exists $attrs->{$attr};


oalders:
Thanks @skaji!

haarg:
This breaks some uses of images: https://metacpan.org/pod/Acme::CPANAuthors::Nonhuman

skaji:
@haarg thanks for pointing out that.

## before
screen shot 2017-05-05 at 14 16 01

## after
screen shot 2017-05-05 at 14 16 27

## what to do

If we keep previous behavior, then I think we should set `class="img-responsive"` only if ``  does not have the `width` attribute.

@haarg What do you think?

haarg:
@skaji We should probably investigate the current uses of inline images more thoroughly.  Having the width and height set seems a little too incidental to use for this purpose.

karenetheridge:
:(

Grinnz:
Since according to bootstrap docs the img-responsive class adds the following styles: `max-width: 100%;, height: auto; and display: block;` then instead of setting that class, the first two styles could be set, excluding `display: block`. That seems to work for scaling the Parallel::Pipes image, and images like in Acme::CPANAuthors::Nonhuman don't each go on their own line.

skaji:
My original intent for this PR is below.

Since #1844, style attributes for `` tags have been not allowed.
I used to use ``, but it does not work anymore.
So I want to add `style="max-width:100%;"` to `` tags somehow.

Yes, I missed the inline usage of `` tags, sorry.

I'm totally okay with reverting this PR for a while.

@Grinnz Thanks. Sounds good.

PR #1876, opened 2017-05-08
Show less is the opposite of show more.

oalders:
The wording of the Changes ("Show more/hide more") was a bit confusing.


Getty:
\o/

PR #1878, opened 2017-05-11
Add 06perms to the front end

oalders:


oalders:
This has integration tests, but no unit tests.  I can add the unit tests if you'd like them.

PR #1889, opened 2017-05-23
convert deprecated query elements

mickeyn:
We have leftover query elements that are deprecated since ES 2.0
I hope this change won't cause issues... at least it doesn't break any test.

Of course, later on all these queries will move to the API... this is an interim cleanup/simplification phase.


oalders:
👍

PR #1894, opened 2017-05-31
Replace activity query with new API endpoint

mickeyn:
API now provides a new endpoint for activity info, we don't
need to send a query from WEB anymore.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Activity.pm
+++ b/lib/MetaCPAN/Web/Controller/Activity.pm
@@ -6,75 +6,18 @@ BEGIN { extends 'MetaCPAN::Web::Controller' }
 
 use DateTime;
 
-my %res = ( week => '1w', month => 'month' );
-
 sub index : Path : Args(0) {
     my ( $self, $c ) = @_;
-    my $req = $c->req;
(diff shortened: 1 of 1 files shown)


lib/MetaCPAN/Web/Controller/Activity.pm
```diff
@@ -6,75 +6,18 @@ BEGIN { extends 'MetaCPAN::Web::Controller' }
...
-    if ( my $module = $req->parameters->{module} ) {
-        push( @$q, { term => { 'dependency.module' => $module } } );
-    }
-    if ( $req->parameters->{f} && $req->parameters->{f} eq 'n' ) {
-        push( @$q,
-            { term  => { first  => 1 } },
-            { terms => { status => [qw< cpan latest >] } },
-        );
-    }
+    my %args = map { $_ => $c->req->parameters->{$_} }
```
mickeyn: @oalders wasn't sure whether to create a new hash with the args or just pass `$c->req->parameters` to `request` - what do you think?

oalders: Good call.  I think that's a good way to handle it.

PR #1906, opened 2017-06-04
Use new /release/versions/DIST API endpoint

mickeyn:
Replace the Elasticsearch query sending with a call to the new
API endpoit.

This change also fixes the related tests and simplifies the
way 'verisons' query was called.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Feed.pm
+++ b/lib/MetaCPAN/Web/Controller/Feed.pm
@@ -136,14 +136,12 @@ sub distribution : Local : Args(1) {
     $c->cdn_max_age('1y');
     $c->add_dist_key($distribution);
 
-    my $data = $c->model('API::Release')->versions($distribution)->recv;
+    my $versions = $c->model('API::Release')->versions($distribution);
+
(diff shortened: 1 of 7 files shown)


haarg:
This seems to take one of the asynchronous calls and makes it blocking instead.  That doesn't seem like a good change.

The API being used for the async calls is not the best.  Maybe we need to change that to make it easier to manage?

mickeyn:
@haarg I noticed that, but in this case it didn't seem to matter.

I agree the way we currently do the async calls is way too complicated and hard to maintain.

In this specific case i took a fast, limited query (size=>250) which was also used in other places where async wasn't needed - so i don't believe the performance is affected in a user-noticeable way.

We can restore the versions call into the async calls, but i do want to clean that code first as you suggested.

mickeyn:
I'm going to revert this commit as I see it's responsible for tests currently breaking in master (odd...)

mickeyn:
@haarg @oalders I reverted this commit and will work on a new one soon
https://github.com/metacpan/metacpan-web/pull/1911

PR #1911, opened 2017-06-08
Revert "Use new /release/versions/DIST API endpoint"

mickeyn:
This reverts commit e6151d0293b89060b3b0ed0e72cbac34a6a725ec.

This commit is causing some tests to fail on 'master', will
investigate.


mickeyn:
it seems that in the last few days we started seeing randomly(?) failing tests with messages like:
t/controller/raw.t ....................... No subtests run 

I think it's random as in some cases restarting the build doesn't run into the same failures.

mickeyn:
i'll go ahead and merge this PR as it's just a revert, no new features

PR #1912, opened 2017-06-11
Switch from AnyEvent to IO::Async

haarg:
We use AnyEvent to run multiple HTTP requests simultaneously.  This converts those to using Net::Async::HTTP.  All of the API model methods have been changed to consistently return futures that can be fetched with `->get`, which should make it easier to clean up how we are using concurrent requests.

Additionally includes a rewrite of the site map generation to something that should actually work, and some other prereq cleanups.


haarg:
I pulled some changes from this PR out into #1915 because they are simpler and don't rely on this.  Will rebase this once that PR is merged.

mickeyn:
@haarg merged

haarg:
The sitemap changes could maybe go in a separate PR, but to handle it sensibly it would have to depend on this one, and require an additional cpanfile.snapshot regeneration.  The existing sitemap code doesn't even work properly, and we aren't running it on the servers.

mickeyn:
it's OK with me (i'm known to combine several issues in single PRs :))

PR #1921, opened 2017-06-15
Remove needless imports

mickeyn:
While moving queries to API calls some of these ES output
formatters are not needed anymore (and will soon be removed
altogether)


mickeyn:
this is a simple, straight-forward cleanup. i'll go ahead and merge it.

PR #1926, opened 2017-06-21
Use /release/versions API endpoint

mickeyn:
This time without breaking it off the async api.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Feed.pm
+++ b/lib/MetaCPAN/Web/Controller/Feed.pm
@@ -14,9 +14,6 @@ use Path::Tiny qw/path/;
 use Text::Markdown qw/markdown/;
 use XML::Feed ();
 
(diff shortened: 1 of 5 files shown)


metacpan_web.conf.local
```diff
@@ -0,0 +1,40 @@
+# rename this file to metacpan::web.yml and put a ':' after 'name' if
```
oalders: Did you intentionally add the `.local` config to the repo?

mickeyn: oh, no :)

mickeyn:
the `pod` test failing is odd... will have to investigate

oalders:
push build is failing

mickeyn:
yeah, that pod test failure is odd... couldn't yet find what's going on, other than the text having a tiny difference.

mickeyn:
seems like the previous pod test failure was random and happened in different branches inc. master.

PR #1933, opened 2017-06-28
Search: fix author search results reading

mickeyn:
After moving to a new API endpoint the structure name has
changed from 'results' to 'authors'.

This will fix the authors matches not showing on the search results page.


mickeyn:
after the tests passed, since this is a no-risk change that will only have a positive effect - i'm going to merge it.

PR #1935, opened 2017-06-29
Dockerize the MetaCPAN

zakame:
This adds a `Dockerfile` so that the app can be built as a Docker image.

    docker build -t metacpan-web .

One can start a development server on Docker via

    docker run -p 5001:5001 -v /path/to/metacpan-web:/metacpan-web -it metacpan-web

and this will be a working setup, using https://fastapi.metacpan.org as its default backend.

This is extracted and rewritten from metacpan-docker to allow independent usage (i.e. one doesn't need to source any other repo in order to build this as a Docker image.)

This is part of an effort to Dockerize MetaCPAN (including the API backend and ElasticSearch services.)  See https://github.com/metacpan/metacpan-docker/pull/11 for more info.


Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+COPY cpanfile cpanfile.snapshot /metacpan-web/
+WORKDIR /metacpan-web
+
+RUN cpanm App::cpm Carton && \
+    useradd -m metacpan-web -g users && \
+    mkdir /carton && \
+    cpm install -L /carton
+
+COPY . /metacpan-web
```
haarg: Is there a reason to do this copy rather than just running directly from the repo?

zakame: Thanks, that's a good question.  In Docker Compose, most environments `COPY` the repo into the resulting image so that is can remain standalone, e.g. it can be put into hub.docker.com or local private Docker repository (like quay.io or gcr.io) and be able to pull/run from there without needing to clone the repo prior, e.g:

    docker pull my-private-docker-repo.io/metacpan-web:1.0
    docker run --rm -p 5001:5001 -it my-private-docker-repo.io/metacpan-web:1.0

In https://github.com/metacpan/metacpan-docker/pull/11 we do mount the repo into the container (via `services.web.volumes` in the Compose configuration,) so the `COPY` can be removed if we prefer to have a leaner "barebones" image of `metacpan-web` instead.  This would mean however that on ordinary `docker` runs we'd have to mount in the repo manually, e.g.:

    docker run -v /path/to/metacpan-web:/metacpan-web -it metacpan-web

haarg: I guess it depends on what the purpose of this is.  It seems like it's primarily for development, so the leaner approach seems better.  The plackup command being used is also for set up for development.  The -r option provides no real value if this isn't pointed at files that are being edited, and if you aren't using -r, you would want to use a different server like Gazelle or Starman.

zakame: Great, I'll remove the `COPY` then.  To add, I am also thinking that since Docker 17.05 has [multi-stage builds][0] now it would be possible to `COPY` the repo in much later, for production-style deployments.

[0]: https://docs.docker.com/engine/userguide/eng-image/multistage-build/

Just to note, the `CMD` can also be overridden at runtime by the user, e.g. via `docker run metacpan-api plackup -S starman` or similar.

oalders:
@zakame what's the status of this pull request?  Is it ready for review?

zakame:
@oalders yes, I've been using it myself to work on -web :)

oalders:
Thanks @zakame!

PR #1943, opened 2017-07-05
Read correct key from output

mickeyn:


mickeyn:
no risk change - merging

PR #1945, opened 2017-07-07
refactor and improve file sorting on release page

haarg:
Move all file sorting into one routine.  Only link one one of a matching
pod/pm pair.  Allow for other interesting files that aren't in the root
directory.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Release.pm
+++ b/lib/MetaCPAN/Web/Controller/Release.pm
@@ -5,8 +5,6 @@ use Moose;
 use namespace::autoclean;
 
 BEGIN { extends 'MetaCPAN::Web::Controller' }
-use List::Util ();
-use Ref::Util qw( is_arrayref );
 
 with qw(
     MetaCPAN::Web::Role::ReleaseInfo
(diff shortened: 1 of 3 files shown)


lib/MetaCPAN/Web/Controller/Release.pm
```diff
@@ -162,54 +134,52 @@ sub view : Private {
...
     };
 
-    my %skip;
-
-    for my $f (@$files) {
-        next if $f->{documentation};
-        my @modules
-            = is_arrayref( $f->{module} ) ? @{ $f->{module} } : $f->{module};
-        for my $module ( grep {defined} @modules ) {
+    for my $f ( @files{@files} ) {
```
mickeyn: :)

mickeyn: isn't this just `values %files`? i don't see any code changing the hash or the array.

haarg: Left over from refactoring where I wanted sorting or removed items from `@files`.  It could just be values in this case.

PR #1946, opened 2017-07-07
Release Info refactor

haarg:
Refactor of release info fetching and collection.  Error handling not yet complete.


lib/MetaCPAN/Web/Controller/Release.pm
```diff
@@ -203,6 +162,103 @@ sub _files_to_categories {
     return $ret;
 }
 
+my $rt_cpan_base = 'https://rt.cpan.org/Ticket/Display.html?id=';
```
oalders: Any reason we didn't want to have all of this logic living in a model?

haarg: It kind of feels like it doesn't belong there, since it is display logic.  I wasn't sure really about where it should be, but ReleaseInfo would be the appropriate model if it was on that level.

haarg: Actually I misread your comment.  I was considering moving this into ReleaseInfo, but I wasn't certain yet if I wanted to do that.  But it can also be addressed separately.

oalders: I was just curious.  I thought it might be easier to test in a model and I think it got pulled from a model into the controller.

PR #1948, opened 2017-07-11
Use new /search/first API endpoint

mickeyn:
Simplify data reading using the new endpoint.


oalders:
Failure on the pr build

mickeyn:
this is odd. t/controller/search/precision.t fails on Travis but not on my VM... I can't recreate the failure.

mickeyn:
looks better now.

PR #1949, opened 2017-07-11
Favorite model: check input args before URL concat

mickeyn:
This will fix some log warnings.


oalders:
Failure on the pr build

mickeyn:
I'll close this PR. if the warnings are still relevant we'll try to deal with them again.

PR #1966, opened 2017-08-30
Use a striped table for the diff file list

ilmari:
This makes it easier to match up the file names and the
insertion/deletion counts, especially on wide screens.
Also make the counts links to the diff chunk.


oalders:
@ilmari do you have a before/after screenshot for this?

ilmari:
@oalders

oalders:
LGTM

jberger:
Looking at those images, the thing I bump on is the offset caused by the double digit diffstat

ilmari:
@jberger That's the same before and after, but that could be fixed by making the two numbers separate table columns. In fact, I initially did that, but that made those two columns much wider, and fixing that is beyond my CSS skills.

jberger:
@ilmari yes, I know they aren't related. I was just making a side comment. It is also beyond my quite limited CSS skillset.

PR #1981, opened 2017-11-08
Reverse dependencies: pass 'sort' in the request body

mickeyn:
The 'sort' param is a complex structure (hash) which gets
stringified if passed as a URL param which leads to an error
on the API side, and data not to be displayed on the page.

This change makes the request to be sent as POST.


mickeyn:
closing in favor of a better solution

PR #1991, opened 2017-11-17
Use grep metacpan for distro search

atoomic:
note that grep.metacpan is only going to perform
the search for the last available version of
the distribution.

Left bar search inside a distro

Results in grep.metacpan

### DIFF
--- a/root/inc/release-tools.html
+++ b/root/inc/release-tools.html
@@ -41,6 +41,14 @@
         <input type="submit" style="display: none">
     </form>
 </li>
+<li>
+    <form action="https://grep.metacpan.org/search">
+        <input type="hidden" name="qd" value="<% release.distribution %>">
+        <input type="hidden" name="source" value="metacpan">
+         <input type="search" name="q" placeholder="grep distribution" style="width: 153px" class="form-control tool-bar-form">
+         <input type="submit" style="display: none">
+     </form>
+</li>
 <%- IF versions.size > 1 %>
 <li>
     <select onchange="document.location.href='/<% IF module; "module"; ELSE; "release"; END %>/'+this.value" class="form-control tool-bar-form">


root/inc/release-tools.html
```diff
@@ -35,8 +35,9 @@
     </button>
 </li>
 <li>
-    <form action="/search">
-        <input type="hidden" name="q" value="dist:<% release.distribution %>">
+    <form action="http://grep.metacpan.org/search">
```
haarg: Should be https.

atoomic: updated to https, was not sure it was available :-)

oalders:
My concern here is that this may end up being confusing for users. To start with, I think I'd prefer having a search in dist + a grep in dist link.  There's room for it in the UI.  The grep link could use the magnifying glass icon to make it consistent.

Consider (bad example) some search like https://metacpan.org/search?q=dist:DateTime-Format-MySQL+datetime%20format%20mysql

Inside the dist, it returns at least one useful result. On the grep app, it will return nothing (if I got the query correct): https://grep.metacpan.org/search?q=datetime+format+mysql&qd=DateTime-Format-MySQL&qft=

So, I think that a full text search won't always translate well to a grep.  It would be nice to have both options at least to start with and people could let us know if we should drop the old one or continue on with it.

atoomic:
thanks for the feedback, I will adjust the [minor] changes in that direction 
I've checked and it appears that grep is right to display no result for "datetime format mysql" as this is not used anywhere in the source code for the DateTime-Format-MySQL distro

atoomic:
@oalders I've updated the PR as suggested. But I've discarded adding the fa search icon to the placeholder, as this requires changing the font-family for this input search and looks awkward when compare side/side

Dual search in Tools

oalders:
Could you rebase this branch and also change `Grep` to lower case?  At that point, I think we're in good shape.  We could add a grep link to the search results as well in a different pull request.

atoomic:
@oalders rebased and change `Grep` to `grep`

oalders:
Thanks! Build failure is unrelated.

PR #1994, opened 2017-11-17
Leo/updates with gazelle

ranguard:


cpanfile.snapshot
```diff
@@ -3394,6 +3434,30 @@ DISTRIBUTIONS
       perl 5.008001
       strict 0
       warnings 0
+  MooseX-Role-WithOverloading-0.17
```
haarg: What did you use to rebuild the snapshot?  This seems like a needless prereq.

PR #1995, opened 2017-11-18
change badge (png -> svg) and add kritika badge

skaji:
same as https://github.com/metacpan/metacpan-api/pull/771


oalders:
Thanks @skaji!

PR #2001, opened 2017-11-19
Modified GH issues box text

mickeyn:
We request 15 issues and display up to 15 issues, but the text
is hard-coded to "Last 15 issues" where there can be less
than 15 returned.


oalders:
👍

PR #2006, opened 2017-12-03
Show the River of CPAN position for distributions

tsibley:
Affects the search results, the release page and all documentation pages
under a release, and all tables of releases (recent page, author page,
reverse dependencies, etc).

Resolves GH#1679.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/River.pm
+++ b/lib/MetaCPAN/Web/Controller/River.pm
@@ -1,0 +1,29 @@
+package MetaCPAN::Web::Controller::River;
+use Moose;
+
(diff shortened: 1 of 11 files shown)


oalders:
Restarted the build as it's not obvious to me how the test failure would be related to this code.

tsibley:
Ah, the failure probably is real. I bet the test is too coupled to the author release table.  I’ll take a look this week.

tsibley:
Repushed with a small correction to the `t/controller/author.t` file that makes it pass locally.

oalders:
Thanks @tsibley!

PR #2008, opened 2017-12-03
pod2html render

haarg:
Render arbitrary Pod to HTML the same way MetaCPAN normally does.  Can aid in debugging and other things.


haarg:
This should be ready now.  In changes the Lab link in the header to a Tools link, which has a nav sidebar with both the lab features and the pod renderer.

The pod renderer accepts pod either via a textarea or a file upload, and does rendering via ajax, with standard form post fallback.  I'd like to add the ability to drag and drop files onto it, but that can wait.

PR #2011, opened 2017-12-11
Clean up river gauge

haarg:
Layout and some other cleanups for river gauge.  It also embeds the svg directly on pod and release pages.  We could also embed it on other pages if that was desired, but the data wasn't immediately available so I didn't both at this point.

The html br tags in the SVG don't work in Firefox, but an encoded newline seems to, so I replaced them.

### Search page
Chrome before:
search-chrome before
After:
 search-chrome after
Firefox before:
search-firefox before
After:
search-firefox after
### Pod breadcrumbs
Chrome before:
pod-chrome before
After:
pod-chrome after
Firefox before:
pod-firefox before
After:
pod-firefox after
### Release list
Chrome before:
author-chrome before
After:
author-chrome after
Firefox before:
author-firefox before
After:
author-firefox after

### DIFF
--- a/root/inc/breadcrumbs.html
+++ b/root/inc/breadcrumbs.html
@@ -48,8 +48,8 @@ END %>
       IF have_released; %>
     <a class="latest" href="<% IF module %>/pod/<% module.documentation %><% ELSE %>/release/<% release.distribution; END %>" title="<%- IF release.maturity == 'developer'; 'dev release, '; END %>go to latest"><span class="fa fa-step-forward"></span></a>
   <%- END; END; %>
-  <div class="inline"><% INCLUDE inc/river-gauge.html, distribution = release.distribution %></div>
-  <div class="inline"><%- INCLUDE inc/favorite.html %></div>
+  <% INCLUDE inc/river-gauge.html, distribution = release.distribution %>
+  <%- INCLUDE inc/favorite.html %>
   <%- IF module %>
   &nbsp;/&nbsp;<span <% IF schema_org %>itemprop="name"<% END %> ><% module.documentation or module.module.0.name %></span>
   <%- END %>
--- a/root/inc/river-gauge.html
+++ b/root/inc/river-gauge.html
@@ -5,6 +5,6 @@
 <object data="/river/gauge/<% distribution | uri | html %>"
   type="image/svg+xml"
   width="24px"
-  height="16px"
+  height="15px"
   alt="river gauge for <% distribution %>"
-  class="river-gauge"></object>
+  class="river-gauge-gauge"></object>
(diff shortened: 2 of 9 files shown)


root/inc/river-gauge.html
```diff
@@ -2,9 +2,9 @@
   Unlike an <img>, an <object> grafts the SVG document into the DOM, which
   means browsers will display the <title> elements of the SVG.  Yay!
 -->
-<object data="/river/gauge/<% distribution | uri | html %>"
+<object data="/river/gauge/<% distribution | uri %>"
```
tsibley: Relying on `uri` to do enough escaping for HTML too is bad practice.  Why did you remove it here?

root/inc/breadcrumbs.html
```diff
@@ -48,8 +48,8 @@
       IF have_released; %>
     <a class="latest" href="<% IF module %>/pod/<% module.documentation %><% ELSE %>/release/<% release.distribution; END %>" title="<%- IF release.maturity == 'developer'; 'dev release, '; END %>go to latest"><span class="fa fa-step-forward"></span></a>
   <%- END; END; %>
-  <div class="inline"><% INCLUDE inc/river-gauge.html, distribution = release.distribution %></div>
-  <div class="inline"><%- INCLUDE inc/favorite.html %></div>
```
tsibley: The reason for removing the `inline` class in an otherwise unrelated change isn't mentioned.  From the face of it, it seems like it would break https://github.com/metacpan/metacpan-web/blob/d3f34a1/root/static/js/cpan.js#L301 (although the correctness and provenance of that line seems dubious at first glance).

tsibley:
> The html br tags in the SVG don't work in Firefox, but an encoded newline seems to, so I replaced them.

Yeah, I realized the `<br>` didn't work in Firefox, but didn't think to encode them.  Nice!

root/inc/breadcrumbs.html
haarg: The div serves no real purpose, and then uses a silly class like "inline" to fix it.  I'll fix the javascript to handle this properly.

tsibley: I totally agree it's a stupid bit of markup. :-)

tsibley:
@haarg Did you see https://github.com/metacpan/metacpan-web/pull/2011#discussion_r156277696?  It looks like GitHub decided that was on an outdated commit for some reason and doesn't show it by default.

root/inc/river-gauge.html
haarg: I was checking on how the default filtering is applied and it slipped into the commit.

tsibley:
Hrm.  Switching to `<svg:title>` appears to have broken the tooltip when the SVG is embedded in the page (on both Chrome and Firefox).  It still works on the SVGs which are included via an `<object>`.

tsibley:
It's a stupid state of affairs, but I guess we can vary the use of `<title>` and `<svg:title>` with the value of `embed`.

tsibley:
(Otherwise, this looks great and I appreciate the cleanups.)

tsibley:
Merged!  Thank you!

PR #2019, opened 2018-01-29
remove issue template

haarg:
I don't see any real purpose to the current issue template.  All it does is ask people to not file issues.  I would rather have people file their requests than not, and would rather that they be created as issues rather than filed on the wishlist to be ignored.


oalders:
I'm fine with removing this.  @ranguard thoughts?

ranguard:
Dont feel strongly - just useful to set expectations of what we are likely to work on

PR #2021, opened 2018-01-29
enable perltidy --weld-nested-containers

haarg:
I'm undecided on this change.

### DIFF
--- a/lib/MetaCPAN/Sitemap.pm
+++ b/lib/MetaCPAN/Sitemap.pm
@@ -62,16 +62,15 @@ sub _request {
         $content      = $json->encode($content);
     }
     $url .= '_search/scroll?scroll=1m&size=' . $self->size;
-    $self->ua->POST( $url, $content, content_type => $content_type, )->then(
-        sub {
-            my $response = shift;
-            my $content  = $json->decode( $response->content );
-            return Future->done
-                if !@{ $content->{hits}{hits} };
-            $cb->( $content->{hits}{hits} );
-            return $self->_request( $content->{_scroll_id}, $cb );
-        }
-    );
+    $self->ua->POST( $url, $content, content_type => $content_type, )
+        ->then( sub {
+        my $response = shift;
+        my $content  = $json->decode( $response->content );
+        return Future->done
+            if !@{ $content->{hits}{hits} };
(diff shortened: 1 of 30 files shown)


lib/MetaCPAN/Sitemap.pm
```diff
@@ -62,16 +62,15 @@ sub _request {
...
-        }
-    );
+    $self->ua->POST( $url, $content, content_type => $content_type, )
+        ->then( sub {
+        my $response = shift;
+        my $content  = $json->decode( $response->content );
+        return Future->done
+            if !@{ $content->{hits}{hits} };
+        $cb->( $content->{hits}{hits} );
+        return $self->_request( $content->{_scroll_id}, $cb );
```
tsibley: This appears to be a mis-indentation, so perhaps a bug in the weld-nested containers option?

lib/MetaCPAN/Web/Controller/Diff.pm
```diff
@@ -12,7 +12,7 @@ sub release : Local : Args(4) {
     my ( $self, $c, @path ) = @_;
     my $diff = $c->model('API::Diff')->releases(@path)->get;
     $c->stash(
-        { diff => $diff, template => 'diff.html', type => 'release' } );
+    { diff => $diff, template => 'diff.html', type => 'release' } );
```
tsibley: Similar misindentation here (but not just in `sub` context).

lib/MetaCPAN/Web/Controller/Release.pm
```diff
@@ -53,16 +53,14 @@ sub view : Private {
...
-        }
-    )->get;
+    my $data = $release_info->else( sub {
+        my $error = shift;
+        return Future->fail($error)
+            if !ref $error;
+        $c->detach('/not_found')
+            if $error->{code} == 404;
+        $c->detach( '/internal_error', $error );
+    } )->get;
```
tsibley: Note that the indentation of this nested `sub` is ok.

tsibley:
Notes above on specific weird cases inconsistent with other changes, which include examples of what I think @mickeyn is talking about.  I wonder if they can be solved with another option (or bug fix)?

Otherwise this looks fine, and I think it's probably an improvement.

PR #2042, opened 2018-04-21
Add "How to Contribute" link and functionality

### LINKED ISSUE #1779: Add a "Contributing" link in the left sidebar if the dist has a Contributing file
To encourage people to read the Contributing file (and add them to their own dists :-), how about adding a link in the left sidebar (near Changes and Homepage)?

zakame:
This implements a link to a release's `CONTRIBUTING` or similar files in its MetaCPAN page:

image

Currently this will link to a new route `/contributing-to/$author/$release` which calls ~a new API implemented in https://github.com/metacpan/metacpan-api/pull/802~ `API::Release->interesting_files` and filters for Contributing/Hacking files, as implemented in https://github.com/metacpan/metacpan-api/pull/803.  If a Contributing file is found, it will use the appropriate renderer for it (e.g. HTML if it is in pod, source if otherwise.) If no file is found, it will show a 404 with a suggestion to contact the author.

Fixes #1779.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/ContributingDoc.pm
+++ b/lib/MetaCPAN/Web/Controller/ContributingDoc.pm
@@ -1,0 +1,60 @@
+package MetaCPAN::Web::Controller::ContributingDoc;
+
+use Moose;
+use namespace::autoclean;
+
+BEGIN { extends 'MetaCPAN::Web::Controller' }
+
+use List::Util ();
+
+sub index : Chained('/') : PathPart('contributing-to') : CaptureArgs(0) {
+}
+
+sub release : Chained('index') : PathPart('') : Args(2) {
+    my ( $self, $c, $author, $release ) = @_;
+
+    # force consistent casing in URLs
+    if ( $author ne uc($author) ) {
+        $c->res->redirect(
+            $c->uri_for( $c->action, [ uc($author), $release ] ), 301 );
+        $c->detach();
+    }
+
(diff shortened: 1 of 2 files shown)


oalders:
@zakame thanks for this.  Could we squash this down into one commit as well?

I know we had a misunderstanding with some of this when I last reviewed it.  I apologize for not communicating more clearly initially and then communicating quite rudely subsequent to that.  I need to do better than that and I hope that you'll accept my apology.

My main concern is that if people are going to contribute, I don't want them to spend a lot of time going down a path that we might not be able to use, because that might discourage them from contributing again in future.  That's what I was trying to express, but in a really terrible way.

zakame:
@oalders no worries :smile_cat: I certainly understand the potential additional effort if we maintain more code, and I appreciate your rigor for ensuring that every change counts for both improving the system and keeping it easy enough to hack on. :+1:

zakame:
@oalders now squashed, I'll do the same for the corresponding -api PR as soon as I get feedback on https://github.com/metacpan/metacpan-api/pull/803#discussion_r186600089 (or sooner, if you prefer me making another PR later to improve on that.)  Thanks!

oalders:
@zakame I think you can go ahead and squash the other one.  If there turns out to be an issue, we can look at dealing with it.  Ideally the API would already know which files are "interesting", so possibly this is something we should be figuring out at indexing time rather than on the fly.  For now, I suspect it'll be fine.

Issue #2055, opened 2018-05-30
plussers page has bad fallback icon


haarg:
The fallback icon on the plussers page is far too resolution for the size it is displayed.  We should either find a better icon to use, or use one of gravatar's fallbacks.

haarg:
Fixed by using identicons

Issue #1623, opened 2015-11-26
Pages of some dists have http content


openstrike:
I've just today noticed that there are some pages (well, one anyway) on https://metacpan.org/ generating warnings in the browser about unencrypted items (Firefox 42.0). The page for [File::Which](https://metacpan.org/pod/File::Which) is one such, although I haven't managed to spot which entity on that page is causing the problem.

I found an old issue (#361) which might be related to this but was closed apparently without total resolution. Sorry I can't be of more help but wanted to make sure that this issue didn't go unrecorded.

oalders:
Is this the problem? `<div class="breadcrumbs" itemscope itemtype="http://schema.org/SoftwareApplication">`

haarg:
This is going to be mostly due to gravatars in the plussers list.

We have functions that will rewrite the URLs to use https where possible.  But users can specify arbitrary URLs for their avatar, which often aren't available via https.  I've just some improvements to the code that rewrites the URLs which will handle the particular case that is showing up on File::Which, but it isn't a full solution.

To fix this properly we'd need to cache the non-gravatar avatars locally.

openstrike:
@oalders Apparently not. That was my initial thought too, but it's present in other pages (eg. [Test::Simple](https://metacpan.org/pod/Test::Simple) where the browser doesn't issue the same warning.

oalders:
I'd also be fine with either forcing everyone to use Gravatars or just not displaying gravatars which aren't available via https.  Allowing people to use something other than Gravatar has been the source of subtle bugs for quite a while now.

openstrike:
ICBW but even if you allow avatars from other, arbitrary https URLs doesn't that just move the problem to https URLs with bad certs (because of unkown issuer / bad expiry / SHA-1 / weak cipher / whatever)? Either mandating Gravatars or local caching sound like winners, though.

moltar:
This one has it: https://metacpan.org/pod/Plack::App::Directory

It's the user pic of one of ++ed users: http://www.kjetil.kjernsmo.net/images/kjetil-square-2015.jpg

haarg:
This should be fixed now.

PR #2068, opened 2018-06-03
only use gravatars based on user emails

### LINKED ISSUE #1022: Can't use gravatar after having a custom avatar URL
Quoting @benkasminbullock on #606:

> there seems to be a bug here that if the avatar is entered and then removed again, the gravatar avatar doesn't come back. Shouldn't it use a checkbox for gravatar and label the input box as an alternative to gravatar?

I had a custom avatar URL, but when I try to remove it, I get no avatar at all, instead of using gravatar. There appears to be no way at all for me to make my account use gravatar (other than manually specifying the right gravatar URL, which is what I currently do).

haarg:
This removes the use of avatars customizable through the account page.
Instead, gravatars based on the users emails is used, falling back to
identicons.

The customizable avatars did not work well for the multiple sizes we
wanted to display them in.  This was also a problem for our fallback
icon.  The user images would often be served through http, not https,
giving mixed mode alerts.  If the user modified their profile, the
generated gravatar URL would be baked in, preventing updates to their
email address from updating the avatar.

Fixes #1022, #1623, #2055.


karenetheridge:
nice - this means that Gravatar::URL is now going to be consistent with what metacpan displays.

Grinnz:
Not sure how you're testing it, but Gravatar::URL for that user's email address gives me http://www.gravatar.com/avatar/629e016f4d4cf66337b150a0287a2236.

karenetheridge:
I was in error when I mentioned Gravatar::URL -- actually the problem (this is in Acme::CPANAuthors::Nonhuman FWIW - my most important distribution ever /s) is that the data returned by https://fastapi.metacpan.org/v1/author/$author still contains the custom avatar URLs, not those fetched via Gravatar::URL.  e.g. https://fastapi.metacpan.org/v1/author/ZDM still contains the custom icon rather than default pixel art.

karenetheridge:
To be more specific: there is an unfinished part of the change covered by this pull request -- the API needs to be updated to be consistent with the new value used by the front end.

haarg:
I'm more inclined to remove the value from the API.

PR #2072, opened 2018-06-04
do not link `source` files in sitemaps - we block in our robots.txt file

ranguard:
One of the warnins from Google is our sitemaps link to `/source/...` which we block with our robots.txt

This is a little hack to stop those being included


haarg:
I couldn't see any warnings specific to the sitemaps about source URLs.  We don't include any source URLs in our sitemaps anyway.

PR #2082, opened 2018-06-19
add gitlab to profiles

mziescha:
I don't know what about the license because of using the logo. But since microsoft bought github I though to add gitlab into profile is a good idea.


haarg:
Can you remove the changes to font-awesome?  We don't want to modify those files locally, and there isn't any need for those classes for this change anyway.

mziescha:
Done sry for changing fa localy

oalders:
Thanks @mziescha!

PR #2091, opened 2018-06-27
use non-version specific source links on latest pages

### LINKED ISSUE #1109: Version-indepdent urls for source (non-pod) files
When we added the /pod controller we expanded the namespace and created version-independent urls (based on distribution instead of release).
We specified (in #176) that we might do the same for non-pod urls for consistency.
#1094 and #1097 have specifically requested that (version-independent links to, for example, META files).

haarg:
When on a latest page using a non-version specific URL, make the source
links use non-versioned forms.

Fixes #1109

### DIFF
--- a/root/pod.html
+++ b/root/pod.html
@@ -26,19 +26,25 @@
       Module version: <span itemprop="softwareVersion"><% documented_module.version | html %></span>
     </li>
     <% END %>
+    <% IF permalinks || release.status != 'latest';
+        source_base = '/source/' _ module.author _ '/' _ module.release;
+      ELSE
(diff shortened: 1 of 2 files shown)


oalders:
```
#   Failed test 'content-disposition attachment with filename set'
#   at t/controller/raw.t line 23.
#          got: undef
#     expected: 'attachment; filename=Moose.pm'
#   Failed test 'cotent-type text/plain'
#   at t/controller/raw.t line 28.
#          got: 'text/html; charset=utf-8'
#     expected: 'text/plain; charset=UTF-8'
# Looks like you failed 2 tests of 9.
t/controller/raw.t ....................... 
```

PR #2100, opened 2018-07-12
Add a subset of SVG tags to the filter_html() 'rules' list.

### LINKED ISSUE #2088: svg stripped out of HTML
The HTML filtering strips out svg content.  We should probably allow it.

Example:
https://metacpan.org/source/JGAMBLE/Algorithm-Networksort-2.02/lib/Algorithm/Networksort.pm#L164

jgamble:
Resolves #2088

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Pod.pm
+++ b/lib/MetaCPAN/Web/Controller/Pod.pm
@@ -270,6 +270,35 @@ sub filter_html {
             tr     => [],
             u      => [],
             ul     => [ { id => qr/^index$/ } ],
+
+            #
+            # SVG tags.
+            #
+            circle  => [qw(id cx cy r style transform)],
+            clippath=> [qw(id clippathunits style transform)],
+            defs    => [qw(id style transform)],
+            ellipse => [qw(id cx cy rx ry style transform)],
+            g       => [qw(id style transform)],
+            line    => [qw(id style transform x1 y1 x2 y2)],
+            marker  => [qw(id markerheight markerunits markerwidth orient refx refy)],
+            mask    => [qw(id height maskunits maskcontentunits style x y width)],
+            lineargradient => [
+                        qw(id gradientunits gradienttransform spreadmethod
+                           x1 x2 y1 y2 xlink:href)
+            ],
+            path    => [qw(id d pathlength style transform)],
+            polygon => [qw(id points style transform)],
+            polyline=> [qw(id points style transform)],
+            radialgradient => [
+                        qw(id gradientunits gradienttransform spreadmethod
+                           cx cy fx fy r xlink:href)
+            ],
+            rect    => [qw(id height style transform x y width)],
+            stop    => [qw(id offset style)],
+            svg     => [qw(id height preserveaspectratio version viewbox
+                           width xmlns xmlns:xlink)],
+            title   => [qw(id style)],
+            use     => [qw(id height transform width x xlink xlink:href y)],
         },
         replace_img => sub {
 


jgamble:
I failed the tidyall.t test. Can anyone explain to me what that entails, as I took care to mimic the existing 'rules' entries.

oalders:
```bash
cpanm Code::TidyAll Perl::Tidy
tidyall lib/MetaCPAN/Web/Controller/Pod.pm
```

That should be enough to get you started, I think. It might want you to install some TidyAll plugins, but I can't remember what the behaviour is.

jgamble:
Slight problem: Code::TidyAll doesn't work on Windows, and for the next couple of months all I have to work on are Windows 7 machines.

So I went in the other direction, and ran Perl::Tidy on the version of Pod.pm submitted here. There are a couple of things I'd like cleared up before submitting the corrected file:

First, Perl::Tidy issued a warning about "Unknown option: wn" (this is in the .pertidyrc file). It doesn't seem to be interpreted by either Perl::Tidy or Code::TidyAll, so I removed it (not a patch, just for my own purposes), and got a successfully tidied output. Is '-wn' something I should concerned about?

Secondly, the output tidied not only the code I put in but also some of the code preceding it (which I did not incorporate into the patch). I'm willing to just submit the new version of the patch, but I'm concerned it would get rejected because of this code, and I don't know how I would check for that.

Thanks.

oalders:
We should "squash and merge" this one.

oalders:
> Slight problem: Code::TidyAll doesn't work on Windows

I hadn't considered that.

> First, Perl::Tidy issued a warning about "Unknown option: wn" 

```bash
$ perltidy --help | grep wn
 -wn     weld nested: combines opening and closing tokens when both are adjacent
```

See https://metacpan.org/changes/distribution/Perl-Tidy#L31

Error should go away if you upgrade Perl::Tidy

Thanks for all of your work on this!

jgamble:
> We should "squash and merge" this one.

Just checking -- is this to me? I know about the [squash and merge option on github](https://blog.github.com/2016-04-01-squash-your-commits/), but if there's something I should be doing I'll need some instructions (sorry).

> Thanks for all of your work on this!

My pleasure.

oalders:
> Just checking -- is this to me?

Sorry, that's just a note to whomever pushes the button on this to use that option.  I was hoping @haarg would have a look at this first.

oalders:
@haarg any reason not to merge this?

haarg:
My general concern with this is that the tags are allowed even outside an embedded svg, where they will have a different impact.

It's probably OK to merge, but I'd definitely prefer to add additional restrictions on where the svg tags can be used.  HTML::Restrict doesn't currently provide any way to do that.

jgamble:
> My general concern with this is that the tags are allowed even outside an embedded svg, where they will have a different impact.

It would be nice for it to perform some sort of parsing check, but it's not doing that for HTML either.

But just see what happens, I put some svg tags without an svg block in a test html file. It rendered just fine and ignored the tags (I'm using Firefox), and behaved no differently from what happens when it finds a table row tag outside a table block.

haarg:
It's not that the tags are ignored, it's that they are just treated as generic tags.  Then they are able to have style attributes applied, which we usually filter out.

jgamble:
Honestly, I don't know. Is the Aug 13th post a real objection, or is it just commentary?

oalders:
@haarg are we ok to merge this or do you want to see some changes?

haarg:
I'm ok with merging it as is, but hopefully we can improve this in the future.

PR #2112, opened 2018-09-19
Better landing for dists missing Contributing file

### LINKED ISSUE #2111: "How to contribute" even the author doesn't know
On the modules it has a link at the right saying "How to contribute". I click on this for my own module but find an empty page.

https://metacpan.org/contributing-to/BKB/Lingua-JA-Moji-0.52

However, even I, the author, don't know what I was supposed to put in the module to make the above not appear. Could someone write just a hint as to what the problem is?

zakame:
This adds a new template that renders a standard message for both module
authors and users about adding a Contributing file, as well as
suggestions for its content, to further encourage everyone to contribute
to the CPAN :)

Fixes #2111.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/ContributingDoc.pm
+++ b/lib/MetaCPAN/Web/Controller/ContributingDoc.pm
@@ -32,6 +32,7 @@ sub release : Chained('index') : PathPart('') : Args(2) {
     $c->forward( 'get', [ $author, $release ] );
 }
 
+# /contributing-to/$name
 sub get : Private {
     my ( $self, $c, @args ) = @_;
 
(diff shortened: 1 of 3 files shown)


oalders:
Is this still a WIP?

zakame:
@oalders yep, I'll revisit later this weekend :muscle:

zakame:
@oalders thanks, I've updated to add more content to the placeholder as well as fix minor issues :+1:

oalders:
Thanks for this @zakame. Could you add a test that uses the new template you added?

zakame:
@oalders thanks, I've added a test here :+1: 

I figured testing the new template might be tricky, so I created a new CPAN dist that explicitly doesn't carry `CONTRIBUTING` to use as a live reference :)  Let me know if there are other ways to test the template (or any other things to change. :muscle:)

oalders:
Thanks @zakame!

zakame:
Thanks too @oalders @neilb! :tada:

PR #2116, opened 2018-10-14
Add definition for Mac::SystemDirectory

ssoriche:
This module is required by File::HomeDir but only when installed on a
Mac. It's an optional prerequisite. Having the definition in the
cpanfile.snapshot allows carton install --deployment to install
Mac::SystemDirectory if required, and has no impact if it is not.


oalders:
👍

PR #2131, opened 2018-11-27
take gravatar emails only from cpan.org addresses

haarg:
Stacking multiple gravatars didn't work, so just use the `@cpan.org` address.


oalders:
@haarg what was the problem with this PR?

oalders:
Oh, I see the commit is in master now.

PR #2135, opened 2018-12-08
add some extra surrogate keys

haarg:


oalders:
This is failing some tests.

PR #2148, opened 2019-02-28
Add TPC::NA advertising on the HomePage

atoomic:
The goal is to promote the Perl Conference,
first in US to replace it for YAPC::EU
on June the 22nd. Before going back to normal
after August 15th.

I will set reminder to update this banner over time.


oalders:
👍

PR #2149, opened 2019-02-28
Move the hint as a button title

atoomic:
At this point we only have a single hint,
move it as a "title" for the "I'm Feeling Lucky"
button.

We can consider restoring the previous logic
once we have more than one hint listed.


oalders:
Thanks @atoomic!

Issue #2142, opened 2019-02-14
Feature request: adding Changes to entry summary in RSS feed.


gugod:
Since the latest entry of Changlog of each distribution is parsed out, it might be a good idea to mix that into each entry -- inside which there's only "abstract" now, which is useful, but not giving away the information about what's new.

Since I still favor RSS/Atom feed in the year of 2019, I'd like to spend some time working on this.... will share my progress here.

Update: Here's my WIP:

- (Web) https://github.com/gugod/metacpan-web/tree/latest-changelog-in-feed
- (API) [PR 885](https://github.com/metacpan/metacpan-api/pull/885) Add a new API: /changes/by_releases

I got it almost the way I wanted it, but it is very inefficient (one API query for each feed item), and the change logs is not formatting nice in my RSS reader (tested with Vienna.app) yet.

oalders:
@gugod sounds like a great idea!

gugod:
@oalders thanks for the endorsement ❤️ 

I managed to get the output correct-ish with an inefficient implementation In this WIP branch https://github.com/gugod/metacpan-web/tree/latest-changelog-in-feed 

I will create PR once I figure out how not to add N API queries but just one.... :)

mohawk2:
@gugod Is this now resolved, or are there further changes you'd like to see in the feed?

gugod:
It is up and running for a while and I am happy with the result. 

I was worrying the efficiency of the implementation but maybe it is not too bad afterall.

mohawk2:
PRs to improve efficiency will always be welcome (if evidence is included :-)!

PR #2157, opened 2019-03-16
Latest changelog in feed

gugod:
Context: https://github.com/metacpan/metacpan-web/issues/2142

With this PR, the output of `/feed/recent` action now contains the latest change log entries for each releases. Here's how it looks like when rendered with [Vienna RSS reader](https://www.vienna-rss.com/)

In order to make it looks nice, the "summary" field of each entry is changed to be HTML. This might break some feed consumers. Should it be preferred, I can make this effective only when a parameter is given.

With the current implemention, when there are N recent releases to render (N is usually 100), metacpan-api is queried N/10 times instead of 1 times. I tried with 1 very long query but metacpan-api wouldn't return N results. 10 seems to be the upper bound. I'm guessig this is due to the default resultset size of Elasticsearch, and some parts from metacpan-api needs to be modified in order to retrieve 100 results in one go.

I'm not sure if this newly introduced latency is "fine" for the current server setup. If not, I can certainly patch metacpan-api to yield larger resultset.

Please take a look, and thanks for your time.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Feed.pm
+++ b/lib/MetaCPAN/Web/Controller/Feed.pm
@@ -26,9 +26,19 @@ sub recent : Chained('feed_index') PathPart Args(0) {
     # Set surrogate key and ttl from here as well
     $c->forward('/recent/index');
 
-    ## TODO: fix this N+1 Query.
-    for my $entry (@{$c->stash->{recent}}) {
-        $entry->{changes} = $c->model('API::Changes')->release_changes([ $entry->{author}, $entry->{name} ])->get;
+    my $changes
+        = $c->model('API::Changes')
+        ->by_releases( [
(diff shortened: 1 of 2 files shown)


haarg:
I have some minor concerns with this PR, but I don't have time this week to properly comment on them.  I can follow up next week.

gugod:
Thank you very much @haarg 

I decided to include a new commit b5566da6b91558f379bc36dd0eb409e4f1d56341 as well as a new PR https://github.com/metacpan/metacpan-api/pull/894 on the API side, in order to reduce the roundtrip count. Let me know what you think... since I don't have insights on how good the performance of API server is, I assumed that it is better if the count of roundtrip is lower. If it does not matter that much, it's definitely OK not to include the new PR.

oalders:
Thanks @gugod and @haarg!

PR #2163, opened 2019-04-23
s/https/http/ on dashboard blogs.perl.org link

trapd00r:
On the [dashboard](https://metacpan.org/lab/dashboard), there's a link to neil bowers writeup on how to write a good abstract. The link is dead because blogs.perl.org doesn't seem to like https.
Should be replaced by
http://blogs.perl.org/users/neilb/2014/07/give-your-modules-a-good-abstract.html


haarg:
Thanks.

PR #2176, opened 2019-05-13
Changes: Autolink "github ###" as a GH issue link

### LINKED ISSUE #2175: Autolink "github ###" as a GH issue link in Changes
Thanks to #907, `GH-###` in a `Changes` file is recognized as an issue link.  I just noticed that @pjcj uses `github ###` in the [`Changes` file for `Devel::Cover`](https://metacpan.org/changes/distribution/Devel-Cover).  Would it be possible to also autolink that format?  It seems to be [here](https://github.com/metacpan/metacpan-web/blob/700748459b3ca2cec64a5c9834943343857dd33b/lib/MetaCPAN/Web/Controller/Release.pm#L269) in `master` as I write this.  Thanks for considering this request!

cxw42:
Fixes #2175.

Accepts `github`, `Github`, and `GitHub`.  Also accepts `gitHub` because that way I could just use `[Gg]it[Hh]ub` rather than having to list the alternatives! :)

Passes tests with `./bin/prove` on my Cygwin install.

Thanks for considering this PR!


haarg:
Looks fine.  The regex is already case insensitive, so the case alternates shouldn't be needed.

cxw42:
@haarg thanks for pointing that out - the downsides of long `s///e` replacement blocks :) .  The tests currently only cover uppercase forms of `GH` and `PR`.  Maybe this is a separate question, but since we are case-insensitive, should there be more tests to cover more case combinations?

oalders:
Thanks @cxw42!

Issue #2174, opened 2019-05-06
Dependency graphs replacement


Grinnz:
As the link to stratopan graphs has been removed (#2106) I developed https://cpandeps.grinnz.com as a replacement. If desired it could be linked to or embedded as the stratopan graphs were. I am not sure how it will behave in an iframe so I can make changes as needed if that approach is taken. The graph for each distribution can be reached as https://cpandeps.grinnz.com?dist=Dist-Name.

oalders:
closed via #2178

PR #2178, opened 2019-05-13
Link to dependency graphs

Grinnz:
Possible implementation of #2174. I am not sure on a couple things so please suggest fixes if needed.

* The distribution name needs to be URI-escaped as you can see on https://metacpan.org/release/Text-Tabs+Wrap. I assume the standard Template::Toolkit uri filter can be used. It probably also should be HTML escaped but I don't know a good test case for that.
* I don't really understand the `<i>` tag part, so I just copied it from the disabled stratopan link. I assume this is for the icon next to the link.
* I don't see a benefit to embedding the graph like the stratopan graphs, over just using a standard link that people can open in a new tab if they want to - the same as the other links that are there.

### DIFF
--- a/root/inc/dependencies.html
+++ b/root/inc/dependencies.html
@@ -29,6 +29,12 @@ FOREACH dep IN deps.sort %>
       <%- END -%>
         <i class="fa fa-share fa-fw black"></i>Reverse dependencies</a>
     </li>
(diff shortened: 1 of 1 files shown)


oalders:
Thanks @Grinnz!

PR #2180, opened 2019-05-14
Link to dependency graph unconditionally

Grinnz:
The dependency graph may be configured to show dependencies from other phases or relationships, so it may be useful even if the runtime requirements are empty.


oalders:
Do you want to update this to delete the old Stratopan link just below your changes?

Grinnz:
So updated.

oalders:
Thanks @Grinnz!

PR #2200, opened 2019-07-11
Mention and link to CPAN on main page

Grinnz:
As an attempt to limit confusion on the relationship between CPAN and MetaCPAN, the main page could succinctly describe MetaCPAN's function and link to the CPAN home page.


oalders:
Sorry, I have no idea how this got missed.

Issue #2191, opened 2019-06-23
Feature request: Support README.md files


djzort:
Frequently distributions are shipping README.md files as the only README.x style file.

It would be excellent for MetaCPAN to be able to consider a README.md to be the  "proper" README file and render it as html.

oalders:
I think this makes sense.

PatrickCronin:
I'd like to take a crack at this.

oalders:
Closed via #2214  This should be deployed in an hour or so.

PR #2214, opened 2019-10-14
Render Markdown source files as HTML

PatrickCronin:
Relating to #2191, this PR seeks to render .md files as HTML instead of raw source. Comments, questions and suggestions welcome. I'm new to Catalyst, so one of the things I'd like to address is the code duplication between `source.html` and `md.html`.

### DIFF
--- a/News.md
+++ b/News.md
@@ -1,3 +1,9 @@
+Title: Rendering .md files
+------------------------------
+Date: 2019-10-18T01:54:00
+
+We're now rendering .md files as HTML instead of as raw code.
+
 Title: The traffic is comming
 ------------------------------
 Date: 2018-12-02T19:30:00
--- a/lib/MetaCPAN/Web/Controller/Pod.pm
+++ b/lib/MetaCPAN/Web/Controller/Pod.pm
@@ -4,9 +4,8 @@ use Moose;
 
 use Encode qw( encode decode DIE_ON_ERR LEAVE_SRC );
 use Future;
-use HTML::Escape qw(escape_html);
-use HTML::Restrict   ();
 use HTML::TokeParser ();
+use MetaCPAN::Web::RenderUtil 'filter_html';
 use Try::Tiny qw( try );
 use URI ();
 
@@ -136,7 +135,7 @@ sub view : Private {
         }
     )->get;
 
-    my $pod_html = $self->filter_html( $pod->{raw}, $data );
+    my $pod_html = filter_html( $pod->{raw}, $data );
 
     my $release = $c->stash->{release};
 
@@ -197,7 +196,7 @@ sub pod2html : Path('/pod2html') {
         'POST'
     )->get->{raw};
 
-    $html = $self->filter_html($html);
+    $html = filter_html($html);
 
     if ( $c->req->parameters->{raw} ) {
         $c->res->content_type('text/html');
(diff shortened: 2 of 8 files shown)


oalders:
This looks really good: http://0.0.0.0:5000/md/OALDERS/HTML-Restrict-v3.0.0/README.md 
This also renders nicely: http://0.0.0.0:5000/md/HMBRAND/Spreadsheet-Read-0.81/CONTRIBUTING.md

lib/MetaCPAN/Web/Controller/Md.pm
```diff
@@ -0,0 +1,87 @@
...
+
+sub content : Private {
+    my ( $self, $c ) = @_;
+
+    my $file = $c->stash->{file};
+    $self->add_cache_headers( $c, $file );
+    $c->res->last_modified( $file->{date} );
+    $c->stash( {
+        file                => $file,
+        source              => filter_html( $file->{content} ),
```
oalders: 👍

lib/MetaCPAN/Web/Controller/Md.pm
```diff
@@ -0,0 +1,87 @@
...
+        template            => 'md.html',
+        suppress_stickeryou => 1,
```
oalders: 👍

News.md
```diff
@@ -1,3 +1,9 @@
+Title: Rendering .md files
+------------------------------
+Date: 2019-10-14T21:57:00
+
+We're now rendering .md files as HTML instead of as raw code.
```
oalders: 👍

root/md.html
```diff
@@ -0,0 +1,51 @@
...
+  <li><% file.slop %> lines of pod</li>
+  <li><% file.stat.size | format_bytes %> bytes</li>
+</ul>
+
+<div class="content">
+<% USE MultiMarkdown(heading_ids => 1) -%>
+<% FILTER multimarkdown %>
+<% source %>
+<% END %>
+</div>
```
oalders: This template seems to resemble `root/source.html` very closely.  I wonder if we could share the same template and just add the filter conditionally?

oalders:
I think we still need to deal with the case of markdown files which don't actually exist: http://0.0.0.0:5000/md/HTML::Restrict

oalders:
@PatrickCronin thanks very much for this.  I think we've very close!  @mickeyn and @haarg any comments?

root/md.html
PatrickCronin: Agreed. Taking it one step farther, do you think that this functionality belongs within the Source controller, just as a special case? Or is the new Md controller appropriate? I've gone back and forth several times now.

oalders: I think the Source controller is probably a good place for this.  I think it could go either way, but putting it in the Source controller likely means less lines of code.

PatrickCronin:
@oalders I added a commit that gets rid of the `Md` controller and template, and moves the action to the `Source` controller and template. In doing so we saved about 150 lines, and also resolved the issue you raised about non-existent files. If it looks good, I'll redo the commit history, remove the WIP and Draft status, and submit for real.

PatrickCronin:
@oalders - can I redo the commit history?

oalders:
Please!

oalders:
Thanks @PatrickCronin! 🎉

djzort:
Amazing work. Great stuff.

djzort:
@PatrickCronin send me a private message and please allow me to send you Sydney Perl Mongers shirt.

djzort:
@oalders perhaps this is news worthy? https://metacpan.org/news

oalders:
> @oalders perhaps this is news worthy? https://metacpan.org/news

It is and an addition to that page is in the merged pull request.  :)

PR #2216, opened 2019-10-20
Filter HTML _after_ Markdown to HTML conversion

PatrickCronin:


oalders:
Thanks @PatrickCronin!

Issue #2218, opened 2019-10-23
"results per page" search issues


532910:
1. Open https://metacpan.org/ and search something, like `Mason::Plugin` for example.
2. Scroll to the end and to go the second page: https://metacpan.org/search?p=2&q=Mason%3A%3APlugin&size=20
3. Scroll to the end and switch to 500 results per page: https://metacpan.org/search?p=2&q=Mason%3A%3APlugin&size=500 -> you'll see no results at all.

mohawk2:
I think the `p` parameter should probably be a `from` parameter. That would decouple what page you're on from how many things to show. That would influence #2223 which is otherwise a good approach.

oalders:
This is no longer an issue because page size is no longer an option.

Issue #1476, opened 2015-03-21
RevDeps: switching to more results per page when on last page can result in "no dependents:


neilbowers:
If you look at the revdeps for a dist with 50 shown per page, and go to the last page, for example:

> https://metacpan.org/requires/distribution/IO-All?p=3

If you click on the <b>100</b> button to get 100 results per page, you get a message:

> No distributions depending on IO-All could be found

Which is slightly misleading. It's because you were on page 3, but when listing 100 per page there is no page 3.

In this situation I think you could either go to the last page (ie the second page after switching to 100/page), or just say that when you switch to a different NPP, you go to page 1.

ranguard:
https://github.com/CPAN-API/metacpan-web/blob/master/root/inc/pager.html#L26 

Change:

        <a href="<% c.req.uri_with( p = page, size = page_size ) %>"><% page_size %></a>

To:

        <a href="<% c.req.uri_with( p = 1, size = page_size ) %>"><% page_size %></a>

Someone just needs to send as a pull request

oalders:
With the pagination UI changes, this should no longer be possible.

PR #2223, opened 2019-10-28
pager.html: stay on the same page

some-perl-nerd:
On a result with 10 per page, the second page starts at 10
On a result with 100 per page, the second page starts at 100

As long as you landed on the page from a search result the link shouldn't take you past the end of the results.

Completely untested fix for #2218


oalders:
Thanks for this.  I wonder why we don't just use a pagination module here?

some-perl-nerd:
The pagination plugin doesn't have support for changing page size.

oalders:
I'm ok with not using a plugin if it makes things easier.

mohawk2:
#1476 is related too.

oalders:
We ended up dealing with this by not allowing arbitrary page sizes. 😬 So, I'm going to close this. Thanks for the work @some-perl-nerd. I'm sorry it took so long.

PR #2226, opened 2019-11-08
highlight modules or dists with ADOPTME or HANDOFF or NEEDHELP

ThisUsedToBeAnEmail:
A notification will display at the top of the page in /release and /pod if ADOPTME, HANDOFF or NEEDHELP is found as an author within the modules permissions.


ThisUsedToBeAnEmail:
I would have written this in JS but as the frontend is template toolkit I followed that direction. I have tried to make it extendable so deprecated can be added. There is one oversight that I am aware of and that is using main_module inside of the template and I should check for pod_file and take the module from there. Also wondering whether it should display on modules/provides that do not have pod.

PR #2235, opened 2019-11-17
Display notification when owner is a special PAUSE author.

### LINKED ISSUE #2234: Display notification when owner is a special PAUSE author - NEEDHELP|ADOPTME|HANDOFF
Currently the logic only looks up the co_maintainers for this author where as it can also be found as 'owner'

ThisUsedToBeAnEmail:
The logic was only looking up in the co_maintainers for an author that was NEEDHELP|ADOPTME|HANDOFF. Now it will first check the owner and then the co_maintainers.

Closes #2234

### DIFF
--- a/lib/MetaCPAN/Web/Model/ReleaseInfo.pm
+++ b/lib/MetaCPAN/Web/Model/ReleaseInfo.pm
@@ -283,12 +283,22 @@ sub normalize_issue_url {
 sub normalize_notification_type {
     my ( $self, $data ) = @_;
     if ( is_hashref($data) ) {
-        if (   is_hashref( $data->{permission} )
-            && is_arrayref( $data->{permission}{co_maintainers} ) )
-        {
-            for ( reverse @{ $data->{permission}{co_maintainers} } ) {
-                if ( $_ =~ m/^(NEEDHELP|ADOPTME|HANDOFF)$/ ) {
-                    return $_;
+        if ( is_hashref( $data->{permission} ) ) {
+            my %special = (
+                NEEDHELP => 1,
+                ADOPTME  => 1,
+                HANDOFF  => 1
+            );
+            if (   $data->{permission}->{owner}
+                && $special{ $data->{permission}->{owner} } )
+            {
+                return $data->{permission}->{owner};
+            }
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Web/Model/ReleaseInfo.pm
```diff
@@ -283,12 +283,22 @@ sub normalize_issue_url {
...
-                if ( $_ =~ m/^(NEEDHELP|ADOPTME|HANDOFF)$/ ) {
-                    return $_;
+        if ( is_hashref( $data->{permission} ) ) {
+            my %special = (
+                NEEDHELP => 1,
+                ADOPTME  => 1,
+                HANDOFF  => 1
+            );
+            if (   $data->{permission}->{owner}
+                && $special{ $data->{permission}->{owner} } )
```
oalders: I think `exists $special{ $data->{permission}->{owner} }` would be the safest way to do this.

lib/MetaCPAN/Web/Model/ReleaseInfo.pm
```diff
@@ -283,12 +283,22 @@ sub normalize_issue_url {
...
+            {
+                return $data->{permission}->{owner};
+            }
+            elsif ( is_arrayref( $data->{permission}{co_maintainers} ) ) {
+                for ( reverse @{ $data->{permission}{co_maintainers} } ) {
+                    if ( $special{$_} ) {
```
oalders: This could use `exists` as well.

lib/MetaCPAN/Web/Model/ReleaseInfo.pm
ThisUsedToBeAnEmail: sure I've added a defined and an exists.

lib/MetaCPAN/Web/Model/ReleaseInfo.pm
ThisUsedToBeAnEmail: Added an exists

oalders:
Thanks for all of your help on this @ThisUsedToBeAnEmail.

PR #2236, opened 2019-11-17
Sync RSS link titles with respective feeds

scop:
Various readers and browser add-ons show the link title in feed
discovery phase, and the title may also end up as the title of a
subscription in a feed reader. A plain "RSS" doesn't look good in
either, and is ambiguous in subscription lists.


oalders:
Thanks @scop!

PR #2238, opened 2019-11-18
mailto link: Email address is ARRAY(...)

### LINKED ISSUE #2237: mailto link: Email address is ARRAY(...)
This is very similar to #2232: On https://metacpan.org/pod/XML::XPath the address in the "Take me over?" box is `ARRAY(0xa1c1a38)`. Looking at the typo fixes in https://github.com/metacpan/metacpan-web/pull/2233/files (`intereset` -> `interested`) that page appears to have the fix for #2232, but still the wrong address.

ThisUsedToBeAnEmail:
The API returns inconsistently either an array or a string for author.email take the following examples:

http://localhost:5000/v1/author/NEILB is indexed as "email" : "neil@bowers.com" (locally)
https://fastapi.metacpan.org/v1/author/NEILB is indexed as "email" : ["neil@bowers.com"]
https://fastapi.metacpan.org/v1/author/REHSACK is  "email" : "rehsack@cpan.org"

So update the template logic to check first if an array email.0

Also first check for asciiname before name, which should prevent breakage in the mailto 'URI' for cases like:
"asciiname" : "Audrey Tang"
"name" : "☻ 唐鳳 ☺"
I don't think any validation exits on the asciiname field to validate it is actually ascii, perhaps this needs to be added as a seperate ticket.

Closes #2237

### DIFF
--- a/root/inc/notification.html
+++ b/root/inc/notification.html
@@ -31,9 +31,9 @@
     END;
     IF NOTIFICATION.has_email;
         NOTIFICATION.mailto = MAILTO(
-            (NOTIFICATION.email || author.email),
+            (NOTIFICATION.email || author.email.0 || author.email),
             (NOTIFICATION.title _ ' <' _ NOTIFICATION.module _ '>') | url,
-            NOTIFICATION.body((NOTIFICATION.name || author.name), NOTIFICATION.module) | url
+            NOTIFICATION.body((NOTIFICATION.name || author.asciiname || author.name), NOTIFICATION.module) | url
         );
     END;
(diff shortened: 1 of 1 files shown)


oalders:
@mickeyn do we need to fix up the author.email field on the back end?

@ThisUsedToBeAnEmail see https://github.com/metacpan/metacpan-api/blob/master/lib/MetaCPAN/Util.pm#L130

mickeyn:
@oalders yes, should be checked in the backend. that's an item for the hackathon.

PR #2250, opened 2019-12-07
Use /var/tmp for tmpdir in docker

atoomic:
When run from a docker container the
local var directory is read only, we
should not use it when run from docker.


atoomic:
related to https://github.com/metacpan/metacpan-docker/pull/48

PR #2252, opened 2019-12-07
Add perl min version to the release page

cees:
This adds the minimum required perl version for this module on the left sidebar of the release page. 

It also upgrades font awesome to the latest version so that we can get a nice >= symbol


oalders:
Closes #2252

PR #2276, opened 2020-03-23
Expand non-Docker hacking instructions

cxw42:
I am not using Docker, and the instructions didn't cover setting up the JS side.  This PR adds installation of JS packages to the README.

I'm using my work computer at the moment (with permission), so I'm adding myself to the `.mailmap` also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,15 @@ this will give you a virtual machine already configured and ready to start devel
 ## Installing manually
 
 If you prefer not to use the Docker, the following commands will get you started:
-commands can be converted to:
 
     $ carton install
+    $ npm install
+    $ export PATH="$(realpath ./node_modules/.bin):$PATH"
     $ ./bin/prove t
     $ carton exec plackup -p 5001 -r
+
+To run the tests in parallel, add `-j8` (or however many CPUs you have) to the
+`prove` command.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -10,12 +10,17 @@ this will give you a virtual machine already configured and ready to start devel
 ## Installing manually
 
 If you prefer not to use the Docker, the following commands will get you started:
-commands can be converted to:
 
     $ carton install
+    $ cpanm -l local WWW::Form::UrlEncoded::XS  # Optional but recommended
```
oalders: Where is this module used?  Rather than adding an extra install step we could just add it to the `cpanfile`.

cxw42: I got a warning when running front-end tests that something was falling back to PP since the XS was missing.  If it's OK to require an XS module for tests, I'll gladly drop it in the cpanfile!

cxw42:
@oalders Would it be worth updating the `cpanfile.snapshot` as well?  I hesitate to do so myself as I am not set up to test the back end.

oalders:
@cxw42 upon merge to master a build artifact with an updated `cpanfile.snapshot` will be created.  We can just use that to update.  I should update the workflow to do this on pull requests as well, but no need to worry about that right now.

PR #2283, opened 2020-03-31
Side bar changes for better clarity and presentation

gordon-fish:
Display "Development release" to side bar (under "Module version")
with matching text color to the release text in bread brumbs area
for dev ("maturity": "developer") releases. This is to make it
more clear that a release (especially when the version only has
an underscore, and no "TRIAL" or other such clues) is in fact a
development release.

When there is only one version for a distribution, display a
disabled drop down with "No other versions" text in place of,
and matching the style of, the "Jump to version" drop down, in
the side bar.

Hide "Latest version" link in side bar when there are no stable
("maturity": "released") releases, which would result in a 404
if followed.

### DIFF
--- a/root/inc/release-tools.html
+++ b/root/inc/release-tools.html
@@ -56,6 +56,12 @@
         <%- PROCESS version_options %>
     </select>
 </li>
+<%- ELSE -%>
+<li>
+    <select class="form-control tool-bar-form" style="font-style: italic" disabled="disabled">
+        <option>No other versions</option>
+    </select>
+</li>
 <%- END %>
 <% IF release.metadata.x_help_wanted %>
 <li class="nav-header">Help Wanted</li>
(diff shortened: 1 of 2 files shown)


root/inc/release-tools.html
```diff
@@ -64,13 +70,20 @@
...
 </ul>
-<%- END %>
+<%- END;
+stable = [];
+FOREACH version IN versions;
+    IF version.maturity == 'released';
+        stable.push(version);
+        NEXT;
+    END;
+END %>
```
oalders: This might be better off as a check to see if there's a release with `release.status == 'latest'` and then short-circuit at that point.  If there's no release with a `latest` status, then the `Latest version` link won't work.

root/pod.html
```diff
@@ -27,6 +27,11 @@
       Module version: <span itemprop="softwareVersion"><% documented_module.version | html %></span>
     </li>
     <% END %>
+    <% IF release.maturity == 'developer' %>
+    <li>
+      <b style="color: #ff4500">Development release</b>
```
oalders: Noted that this will be addressed in a different PR.

root/inc/release-tools.html
```diff
@@ -56,6 +56,12 @@
         <%- PROCESS version_options %>
     </select>
 </li>
+<%- ELSE -%>
+<li>
+    <select class="form-control tool-bar-form" style="font-style: italic" disabled="disabled">
+        <option>No other versions</option>
```
oalders: 👍

root/inc/release-tools.html
gordon-fish: Good point. I'll make that adjustment. Thanks.

gordon-fish: I've updated the pull request with that change.

gordon-fish: https://github.com/metacpan/metacpan-web/blob/377054534562a2989d1cc48dd125b35a9273644b/root/inc/release-tools.html#L73-L80

root/pod.html
gordon-fish: Done :) #2284

PR #2284, opened 2020-04-01
Moved color-related 'release' css-rules from breadcrumbs.less to global.less…

gordon-fish:
… to make it easier to use in other places without unnecessary duplication.

Used this for the new "Development release" text in the side bar in pod.html so that it is always in sync with the breadcrumb color (for dev releases) and thus easier to maintain.


oalders:
Thanks @gordon-fish!

PR #2289, opened 2020-04-15
Upgrade some JS deps

oalders:


mohawk2:
It looks like this is obsoleted by #2305 and might be closed?

oalders:
Good call. Thanks!

Issue #2288, opened 2020-04-14
"Jump to version" missing a version


Ovid:
[I filed a bug with Config::Any](https://rt.cpan.org/Ticket/Display.html?id=132342) because it references Config::General v2.47 and that was causing an install issue on my system (carton said it couldn't find that version). I mistakenly assumed the version didn't exist because when you visit the [Config::General](https://metacpan.org/pod/Config::General) page and click on the "Jump to version" drop down, version 2.47 does not appear.

However, [that version does exist](https://metacpan.org/release/TLINDEN/Config-General-2.47).

Note: this does not appear to be related to #2201 (which is open), but could be a regression on #1565 (which is closed).

oalders:
Weird.  It looks like it's one of the many listed `2.45` versions.

<img width="387" alt="Screen Shot 2020-04-14 at 11 59 07 AM" src="https://user-images.githubusercontent.com/96205/79246608-688bed80-7e47-11ea-8ea2-09db18088098.png">

oalders:
Possibly an issue with `META.yml`.  https://metacpan.org/source/TLINDEN/Config-General-2.47/META.yml

Grinnz:
Perhaps metacpan should try to show the "version" from the tarball name if it differs from META, since distribution version doesn't really mean anything anyway.

oalders:
@mickeyn, thoughts?

mickeyn:
I don't mind changing. will look into it.

haarg:
This is a duplicate of #1527.

We could just ignore the version from META and use the filename instead.  I suspect that would be more accurate overall, but I'm not certain.

Grinnz:
I think it would be reasonable to do, since the filename is the version people see, and the only one with any sort of constraint (filename uniqueness).

oalders:
Closed via #2290

PR #2290, opened 2020-04-15
Fix for version display in version jump list; uses version from download field.

gordon-fish:
Addresses issue #2288


gordon-fish:
Also fixes the similar version lists in the "Diff with version" dropdown, and also the dropdown at the top in the breadcrumb section (thanks, @Grinnz)

oalders:
Thanks, @gordon-fish!

PR #2295, opened 2020-04-29
Add "Developer release" sidebar label to Release, Browse, and Source pages …

gordon-fish:
… to match the Pod page.

Also moved the "Developer release" sidebar label in the `pod.html` a little higher, just below the breadcrumb line so it matches uniformly on all of those pages.

This is related to PR #2283.

### DIFF
--- a/root/browse.html
+++ b/root/browse.html
@@ -10,6 +10,7 @@
   <li class="visible-xs">
     <% INCLUDE mobile/toolbar-search-form.html %>
   </li>
+  <% INCLUDE inc/release-status.html maturity = maturity, banner_class = 'release-banner-source' %>
   <li class="nav-header">Tools</li>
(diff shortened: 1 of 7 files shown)


root/browse.html
```diff
@@ -7,6 +7,13 @@
 </div>
 
 <ul class="nav-list slidepanel">
+  <% IF maturity == 'developer' %>
+  <li>
+    <span class="dist-release maturity-developer">
+      <b class="release-name">Development release</b>
+    </span>
+  </li>
+  <% END %>
```
oalders, line 16: It looks like the `<li>` and its contents don't change in the various use cases.  Could we use an include for that?

gordon-fish: That's a good idea. I'm on it.

gordon-fish: As requested, I made a new template in inc/ to reduce the redundancies.

oalders:
Thanks @gordon-fish!

PR #2296, opened 2020-04-30
Fix some errors in the javascript

haarg:
Fixes a couple issues with the javascript, including having Reverse dependencies override a user trying to cmd+click to open a new tab, and sortable tables often being broken.


oalders:
Some of the failures here are expected, but I think:

```
#  Failed test 'reverse deps link uses distribution name'
#   at t/controller/shared/release-info.t line 189.
```

is new.

haarg:
Should be fixed now.  Remaining failures are unrelated to this branch.

PR #2301, opened 2020-05-17
Improve mobile interface

hstejas:
I have tested a few pages on desktop and android(mobile) with Firefox, and Firefox nightly.
expanded


oalders:
Thanks for this @hstejas! Test failures are unrelated.

Issue #2300, opened 2020-05-09
It is extremely hard to press this button on iOS phones


akarelas:
This button that would close the sidebar is very hard to press (doesn't react usually), so if you want to close the side-bar you are better off hitting Safari's reload icon to save your time.

![IMG_0087](https://user-images.githubusercontent.com/423822/81464018-006ec400-91c7-11ea-8f63-63c75f66abc4.jpg)

hstejas:
This seems to be a general issue on phones, not just iOS.

akarelas:
An idea (taken from the way that the iOS settings UI works) for consideration, could be to close the sidebar whenever you touch ANYWHERE outside that sidebar.

oalders:
I don't think we've ever put a lot of effort into how this looks and works on mobile.  We would be very happy for someone to take this issue on.

hstejas:
https://github.com/metacpan/metacpan-web/pull/2301#issue-419130110 should have addressed this

akarelas:
@oalders Is the HTML source of the page that appears on phones identical to that of desktop browsers? (I cannot do "view source" on mobile, so I can't check myself)

oalders:
> @oalders Is the HTML source of the page that appears on phones identical to that of desktop browsers? 

Yes, all mobile optimizations happen on the client side.

akarelas:
The website appears to have changed, but the issue still persists, as now the "hide" icon seems to have disappeared entirely. The following video was captured on a mobile device: https://files.karelcom.net/public/ca7tdsgqw.mp4

oalders:
Is this still a problem?

akarelas:
Works well.

You could further improve by animating all subsequent show/hide effects (right now only the first `show` effect is animated, all others are instantaneous), but ok :D

oalders:
If someone wants to handle the animation, we can do that.  For now I will close this ticket.  Thanks for letting us know!

PR #2303, opened 2020-05-21
hide slidepanel when you click outside it, on phone devices

akarelas:
Just my attempt to solve #2300

Feel free to discard if it doesn't tie in well with your idea of how the website should behave


hstejas:
https://github.com/metacpan/metacpan-web/pull/2304 should fully fix my previous attempt, not sure which is preferable

PR #2304, opened 2020-05-23
move top-margin to nav-list

hstejas:


haarg:
What is this intending to fix? It causes an empty white space block at the top of the sidebar on desktop. If I revert this, the sidebar still seems to appear properly on narrow displays.

haarg:
I was incorrect about this move causing the empty space. It's not the move, but the margin in general. But since I'm trying to fix that, it would still be goo to understand what this was fixing so I don't introduce any regressions.

hstejas:
@haarg I tried to move the sidebar 30px below so that it doesnt cover the navbar at the top which has the burger show and close buttons

PR #2313, opened 2020-06-16
CI tests partial fix

mohawk2:
The CI has been failing since April or so. Partly that's due to a dep not being there (which this fixes), partly because the search has improved and putting `">` on the end of a search string doesn't fail now.

The rest is because the MetaCPAN backend API no longer returns results from https://fastapi.metacpan.org/author/search?q=author:rjbs%20app&from=0, which the -web tests depend on. I am looking at fixing that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ This will give you a virtual machine already configured and ready to start devel
 If you prefer not to use Docker, the following commands will get you started:
 
     $ carton install
-    $ npm install
+    $ yarn install
     $ export PATH="$(realpath ./node_modules/.bin):$PATH"
     $ ./bin/prove t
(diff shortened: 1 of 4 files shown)


oalders:
Thanks so much @mohawk2. We just need some tidying on this.

mohawk2:
Thanks for the kind words! I know you said on IRC to consider marking the failing API-related tests as "TODO", but it seems to me like these could be merged right now?

oalders:
Yes, these can be merged, but it might be helpful for `t/tidyall.t` to be passing first. `tidyall path/to/file` *should* fix any issues.

mohawk2:
Sorry for the delay. Fixed the `tidyall` issue!

PR #2316, opened 2020-06-18
defer loading javascript

haarg:
Our javascript code can all be deferred until after DOM parsing, but the way some of the code was written prevented that.

Convert all of the inline scripts to use the DOMContentLoaded event directly, so they don't need jQuery at parse time, and defer the external scripts.


root/inc/dependencies-graph.html
```diff
@@ -12,11 +12,13 @@ <h3 class="text-center"><% release.name %></h3>
...
-        iframe.attr("src") || iframe.attr("src", src);
-        if (!iframe) return e.preventDefault();
+    document.addEventListener('DOMContentLoaded', function () {
+        $('#dependencies-graph').on('show.bs.modal', function() {
+            var iframe = $("iframe", this);
+            var base   = "//widgets.stratopan.com/wheel";
+            var src    = base + "?q=<% release.name | uri %>";
+            iframe.attr("src") || iframe.attr("src", src);
+            if (!iframe) return e.preventDefault();
+        });
```
oalders, line 22: We've switched the dep graph over to https://cpandeps.grinnz.com/?dist=Moose, so we can probably just remove this?

PR #2323, opened 2020-06-23
use XML::FeedPP for feed generation

haarg:
XML::Feed is heavier, has some odd bugs, and has prerequisites that are currently broken.


lib/MetaCPAN/Web/Controller/Feed.pm
```diff
@@ -213,38 +211,36 @@ sub build_entry {
...
-            entry  => $entry,
-            host   => $params{host},
-            format => $format,
+        $feed->add_item( $self->build_entry(
+            class => $feed->item_class,
+            entry => $entry,
+            host  => $params{host},
         ) );
     }
+    warn Data::Dumper::Dumper($feed);
```
mohawk2, line 243: Debugging artifact? (the CI failed because of tidyall, btw)

Issue #2309, opened 2020-06-15
webperf - "performance opportunities" per Lighthouse


mohawk2:
https://webpagetest.org/lighthouse.php?test=200615_WM_8d77c5a1320cf7af7badb20f6f423515&run=1 shows these opportunities to improve:
- [ ] Consider using `<link rel=preload>` to prioritize fetching resources that are currently requested later in page load (on the various woff2 resources) - save 1.68s
- [ ] Resources are blocking the first paint of your page. Consider delivering critical JS/CSS inline and deferring all non-critical JS/styles (the generated JS and CSS asset files) - save 1.17s
- [ ] Remove unused JavaScript to reduce bytes consumed by network activity (it's saying Google Analytics and half of the JS asset file are unused) - save 0.7s
- [ ] Consider adding `preconnect` or `dns-prefetch` resource hints to establish early connections to important third-party origins (https://www.google-analytics.com, https://stats.g.doubleclick.net) - save 0.55s
- [ ] Remove dead rules from stylesheets and defer the loading of CSS not used for above-the-fold content to reduce unnecessary bytes consumed by network activity (says ~90% of the CSS asset is unused) - save 0.52s
- [ ] Image formats like JPEG 2000, JPEG XR, and WebP often provide better compression than PNG or JPEG, which means faster downloads and less data consumption (sponsors/yellowbot.png, sponsors/liquidweb_color.png) - save 0.52s
- [ ] also says sponsors/yellowbot.png could be "properly sized"
- [ ] Minifying CSS files can reduce network payload sizes (the generated CSS asset) - save 0.17s

mohawk2:
To capture what the overall test (https://webpagetest.org/result/200615_WM_8d77c5a1320cf7af7badb20f6f423515/) said, in case it goes away:

Lighthouse Perf was 58:

Metric | value
--- | ---
First Contentful Paint | 2.4 s
Speed Index | 2.5 s
Largest Contentful Paint | 3.6 s
Time to Interactive | 2.4 s
Total Blocking Time | 0 ms
Cumulative Layout Shift | 0

mohawk2:
[...]

PR #2326, opened 2020-06-25
Preload

mohawk2:
Partly address #2309 to preload / preconnect.

The `woff2` is hardcoded because browsers that support `preload` appear a subset of the ones supporting `woff2`, so it will never be a waste to preload the `woff2`. There may be a slightly better way, but I cannot see it!

### DIFF
--- a/root/wrapper.html
+++ b/root/wrapper.html
@@ -80,6 +80,9 @@
         <%- IF need_less %>
         <script src="/static/js/less.min.js" type="text/javascript"></script>
         <%- END %>
+        <link rel="preload" href="/static/fonts/fa-regular-400.woff2" as="font" type="font/woff2" crossorigin />
+        <link rel="preload" href="/static/fonts/fa-brands-400.woff2" as="font" type="font/woff2" crossorigin />
+        <link rel="preload" href="/static/fonts/fa-solid-900.woff2" as="font" type="font/woff2" crossorigin />
         <script>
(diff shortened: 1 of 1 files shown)


root/wrapper.html
```diff
@@ -79,7 +79,9 @@
         <%- END %>
         <%- IF need_less %>
         <script src="/static/js/less.min.js" type="text/javascript"></script>
+        <link rel="preload" href="/static/fonts/fa-regular-400.woff2" as="font" />
         <%- END %>
+        <link rel="preconnect" href="//www.google-analytics.com" />
```
haarg, line 84: I don't see any value in adding this. The google analytics script is set up immediately after this, and it's async so it doesn't block any other part of the page.

root/wrapper.html
```diff
@@ -79,7 +79,9 @@
         <%- END %>
         <%- IF need_less %>
         <script src="/static/js/less.min.js" type="text/javascript"></script>
+        <link rel="preload" href="/static/fonts/fa-regular-400.woff2" as="font" />
```
haarg, line 82: This shouldn't be inside this if block. I think it also needs the crossorigin attribute to work properly.

root/wrapper.html
mohawk2: Agreed.

root/wrapper.html
mohawk2: Thanks, addressed both.

Issue #2311, opened 2020-06-15
webperf - "Website scanner for JavaScript vulnerabilities and security headers"


mohawk2:
https://snyk.io/test/website-scanner/?test=200615_WM_8d77c5a1320cf7af7badb20f6f423515&utm_medium=referral&utm_source=webpagetest&utm_campaign=website-scanner shows:
- [x] HIGH: no Strict Transport Security
- [x] HIGH: no Content Security Policy - A computer security standard introduced to prevent cross-site scripting (XSS), clickjacking and other code injection attacks resulting from execution of malicious content in the trusted web page context
- [x] MED: jQuery 1.11.1 is vulnerable
- [x] MED: no X Frame Options - Clickjacking protection: deny - no rendering within a frame, sameorigin - no rendering if origin mismatch, allow-from - allow from specified location, allowall - non-standard, allow from any location
- [ ] LOW: X XSS Protection - A Cross-site scripting filter

mohawk2:
External sites utilised that would need considering for a Content Security Policy:
- https://www.google-analytics.com
- https://api.github.com
- https://www.gravatar.com
- https://st.aticpan.org (not sure)

mohawk2:
With the policy proposed in #2327, the last 3 become irrelevant as I am not proposing to restrict images or styles, and the GitHub API is a CORS concern, not a CSP one.

However, the above list does miss this JS sources:
- https://www.google.com/recaptcha/api.js

mohawk2:
The high and medium priority items are sorted. I am therefore closing this.

PR #2327, opened 2020-06-25
Add a Content-Security-Policy allowing JS only from GA and us

mohawk2:
Partly address #2311 by adding a CSP.

It doesn't restrict images, and only restricts JavaScript so as to mitigate the risk of XSS via MetaCPAN, since we use SSO.

### DIFF
--- a/app.psgi
+++ b/app.psgi
@@ -121,6 +121,27 @@ builder {
         };
     };
     enable sub {
+
+        # put all security-related headers here
+        my $app = shift;
+        sub {
+            my ($env) = @_;
+            Plack::Util::response_cb(
+                $app->($env),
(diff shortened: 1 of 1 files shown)


mohawk2:
This needs turning from ad-hoc code into a Plack middleware, and needs to also allow https://www.google.com/recaptcha/api.js

mohawk2:
This is now in good shape and ready for review.

mohawk2:
I felt that passing a data-structure to a named bit of middleware was better than a bit of code that just mutated an input. However, I have now switched back to just a code-ref that adds headers, since I agree that when one knows Plack, that is sufficiently clear.

PR #2339, opened 2020-07-02
Tweaks to improve Lighthouse score

mohawk2:
Per https://www.webpagetest.org/result/200701_40_82038f60f5c208b84dea00edd39a09c5/ there was still room for improvement:
- Performance 91
- Accessibility 60
- Best practices 77
- SEO 82

This aims to improve most of those.

### DIFF
--- a/root/inc/author-pic.html
+++ b/root/inc/author-pic.html
@@ -24,7 +24,7 @@
 
 <div class="author-pic">
 <a href="/author/<% author.pauseid %>">
-  <img src="<% author.gravatar_image(130) %>">
+  <img src="<% author.gravatar_image(130) %>" alt="Author image">
 </a>
 <strong>
   <a rel="author" href="/author/<% author.pauseid %>"><% author.pauseid %></a>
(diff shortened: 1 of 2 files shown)


root/static/less/font-awesome/_variables.less
```diff
@@ -3,7 +3,7 @@
 
 @fa-font-path:         "/static/fonts";
 @fa-font-size-base:    16px;
-@fa-font-display:      auto;
+@fa-font-display:      swap;
```
haarg, line 6: I don't think this is appropriate. For normal text, swap makes sense because it can be rendered with a fallback font. Font-awesome is kind of a hack, and using custom characters to render icons. There is no fallback font that would ever make sense to use.

mohawk2: Good call, removed. Is there any way to use `font-display: swap` for the actual text, though? Lighthouse is right that the text jumps in about 1500ms after request starts.

haarg: We aren't using a custom font for the actual text, so I'm not sure what the purpose of font-display: swap would be.

mohawk2: Makes sense! Resolving this.

oalders:
Thanks @mohawk2!

PR #2344, opened 2020-07-04
Expand the "notification" addressee list for ADOPTME etc

### LINKED ISSUE #2268: HANDOFF banner should be more verbose
When HANDOFF has permission for a distribution/module, this banner text is added:

> Take me over?
> The maintainer of this distribution is looking for someone to take over! If you're interested then please contact them via email.

Could that mailto: link also include the address of the last maintainer (if different from the first-come permission holder) -- or even include all the comaint holders in cc? I've had a few distributions taken over without my knowledge where I released it last but only have comaint rather than first-come, and I'd have been happy to continue maintaining the distribution (and incorporating the new arrival's patches) if only I had known.

I would also include explicit language in the body of the email noting what user did the last release, and the date of that release -- as often the first-come holder has lost track of what has happened with the distribution since his last release and other people took over, and this saves him the effort of having to look that up. Also it would be good to linkify the module name rather than just including it in plain text.

How about:

    - Hey $first_come_name,
    -
    - I'm interested in taking over the $module_name module. If you can grant me the privilege I will release the next version.

    + Hey $first_come_name (and co-maintainers $name, $name, $name),
    +
    + I would like to take over the <a href="https://metacpan.org/pod/$module_name">$module_name</a> module, last
    released on $date by $releaser_name. If you can grant me the privilege I will release the next version.

This would at least let other comaintainers have a chance to say "wait a minute!".

### LINKED ISSUE #2293: HANDOFF gives email to last releaser, not necessarily first-come
I saw this on `Event` earlier. The then first-come, JPRIT, had added HANDOFF as a co-maint. The email link was to the last releaser (me), who couldn't do anything about it, wasting the emailer's time.

Comments / caveats:
- This isn't visible now, since then I've become first-come and removed `HANDOFF`.
- The HANDOFF-adding happened without me knowing, which is probably a missing PAUSE feature, not a MetaCPAN matter. It seems like there should be a pushed notification.

 #2268 says the banner isn't verbose enough, but by my reading it was Ok. I think (similar to that issue) the addresses mailed to should be `To` the first-come, `Cc` at least the last releaser, and probably all co-maints.

mohawk2:
~~This suppresses `writev` warnings on `EPIPE`.~~ But mainly it addresses, and therefore hopefully fix #2268 and fix #2293. @karenetheridge Do you agree?

EDIT: remove the `writev` stuff in favour of #2346

### DIFF
--- a/lib/MetaCPAN/Web/Model/API/Permission.pm
+++ b/lib/MetaCPAN/Web/Model/API/Permission.pm
@@ -88,7 +88,8 @@ sub _permissions_to_notification {
 
     my %data;
     if ($type) {
-        $data{type} = $type;
+        $data{type}        = $type;
+        $data{module_name} = $data->{module_name};
     }
     Future->done( {
         took         => $data->{took} || 0,
--- a/lib/MetaCPAN/Web/Model/API/Permission.pm
+++ b/lib/MetaCPAN/Web/Model/API/Permission.pm
@@ -86,9 +86,13 @@ sub _permissions_to_notification {
         }
     }
 
+    my %data;
+    if ($type) {
+        $data{type} = $type;
+    }
     Future->done( {
         took         => $data->{took} || 0,
-        notification => $type,
+        notification => ( %data ? \%data : undef ),
     } );
 }
 
--- a/lib/MetaCPAN/Web/Model/API/Permission.pm
+++ b/lib/MetaCPAN/Web/Model/API/Permission.pm
@@ -13,6 +13,17 @@ MetaCPAN::Web::Model::Permission - Catalyst Model for 06perms
 
 =cut
 
+# copied from Model::ReleaseInfo
+my %models = ( _author => 'API::Author', );
+has [ keys %models ] => ( is => 'ro' );
+
+sub ACCEPT_CONTEXT {
+    my ( $self, $c ) = @_;
+    $self->new( %$self,
+        ( map +( $_ => $c->model( $models{$_} ) ), keys %models ),
+    );
+}
+
 sub get {
     my ( $self, $type, $name ) = @_;
     return Future->done(undef) unless $name;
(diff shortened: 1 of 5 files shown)


karenetheridge:
Am I inferring correctly that this PR changes the people in the mailto: link to everyone who has comaint, not just the first-come holder?

mohawk2:
@karenetheridge Yes, that's what I was trying to achieve?

lib/MetaCPAN/Web/Model/API/Permission.pm
```diff
@@ -89,6 +89,7 @@ sub _permissions_to_notification {
     Future->done( {
         took         => $data->{took} || 0,
         notification => $type,
+        notifiable   => [ grep !$special{$_}, @perm_holders ],
```
haarg, line 92: This shouldn't be a separate item.  The `notification` value should be changed to a hash with type, email, and module elements.

lib/MetaCPAN/Web/Model/ReleaseInfo.pm
```diff
@@ -133,37 +133,54 @@ sub normalize {
...
-                    modules => $data->{modules}{modules},
-                    changes => $data->{changes}{changes},
-                    )
-                : ()
-            ),
+        my $n_data = $data->{notification};
+        my $n_future;
+        if ( $n_data->{notification} ) {
+            $n_future
+                = $self->_author->get_multiple( @{ $n_data->{notifiable} } )
```
haarg, line 140: Any logic like this should move into the API::Permissions get_notification_type method.

root/inc/notification.html
```diff
@@ -4,38 +4,36 @@
...
-            body   = format("Hey %s,\r\n\r\nI'm interested in taking over the %s module. If you can grant me the privilege I will release the next version.\r\n\r\n Kind Regards,")
+            body   = format("Hello,\r\n\r\nI'm interested in taking over the %s module. If you can grant me the privilege I will release the next version.\r\n\r\n Kind Regards,")
         },
     }.$notification;
+    IF NOTIFICATION.email;
+        notifiable.push(NOTIFICATION.email);
+    END;
     IF pod_file && pod_file.module.0;
-        NOTIFICATION.module = pod_file.module.0.name;
+        NOTIFICATION.module = pod_file.module.0.name || release.main_module;
```
haarg, line 28: This should be pulling from the notification hash, as described in my other comment.

mohawk2: Now implemented

lib/MetaCPAN/Web/Model/API/Permission.pm
mohawk2: This is done.

lib/MetaCPAN/Web/Model/ReleaseInfo.pm
mohawk2: Now implemented, and ready for re-review!

mohawk2:
`t/model/permission.t` now provides a pattern for how to mock the API. This could probably be generalised a bit more.

PR #2345, opened 2020-07-04
Make tolerant of diverse Perl version-spec

mohawk2:
https://metacpan.org/pod/YATT::Lite::XHF blows up as it has a runtime Perl requirement of: `>= v5.10.1, != 5.17, != v5.19.3`. This fixes that.


mohawk2:
Also suppresses warnings caused by looking at https://metacpan.org/contributing-to/SUNNYP/Captcha-reCAPTCHA-0.98

PR #2346, opened 2020-07-04
filter out warnings from gazelle on client connection close

haarg:


haarg:
This is an idea for an alternative to 9e0f386219056b9279b3f9b8eafd2221930e7f76 as way of filtering out warnings from Gazelle.

Issue #2286, opened 2020-04-05
Audit error logs and create issues


oalders:
`var/log/starman/metacpan-web/starman_error.log`

The logs are full of various template errors etc.  We should try to clean some of these up so that the logs become more useful.

mohawk2:
Of the smaller -web error logs provided earlier (thanks @oalders!), this came up 39 times:

```
[2020/07/03 11:37:08] [ERROR] [/search?q="><script%20>alert(String.fromCharCode(88,83,83))</script>] Caught exception in MetaCPAN::Web::Controller::Search->index "Timed out at /home/metacpan/metacpan-web/lib/MetaCPAN/Web/Controller/Search.pm line 78."
```

The search does give results, but it's quite slow. This probably wants investigating.

mohawk2:
Of the 545 lines of logs, fully 171 are like:

```
[2020/07/03 11:19:50] [WARN] [/release/ETHER/Moose-2.1800] failed to writev: -1 errno:32 at /home/metacpan/carton/metacpan-web/lib/perl5/x86_64-linux/Plack/Handler/Gazelle.pm line 255.
```

That is Gazelle getting an `EPIPE` (32) when trying to write output.

mohawk2:
A further 39 lines were caused by the issue fixed in #2345.

mohawk2:
Nearly all the remaining messages are Unicode-related warnings.

PR #2349, opened 2020-07-12
More log tidying

mohawk2:
Following on from #2286, this PR is intended to suppress all the errors in the most recent log I've seen.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Source.pm
+++ b/lib/MetaCPAN/Web/Controller/Source.pm
@@ -26,6 +26,7 @@ sub index : Path : Args {
             $c->model('API::Module')->get(@module),
         );
     }
+    die "Bad module supplied\n" if grep $_->{code} > 399, $source, $module;
     if ( $module->{directory} ) {
         my $files = $c->model('API::File')->dir(@module)->get;
 
--- a/lib/MetaCPAN/Web/View/HTML.pm
+++ b/lib/MetaCPAN/Web/View/HTML.pm
@@ -91,18 +91,18 @@ Template::Alloy->define_vmethod(
     'text',
     decode_punycode => sub {
         my $url_string = shift;
-        my $uri        = URI->new($url_string);
-        if ( !$uri->scheme ) {
-            $uri = URI->new("http://$url_string")
-                ;    # default to http:// if no scheme in original...
-        }
-
-        # This might fail if somebody adds xmpp:foo@bar.com for example.
-        my $host = $uri->host && eval { $uri->ihost };
-        if ( !$host ) {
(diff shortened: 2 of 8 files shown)


lib/MetaCPAN/Web/View/HTML.pm
```diff
@@ -91,18 +91,18 @@ Template::Alloy->define_vmethod(
...
-
-        # This might fail if somebody adds xmpp:foo@bar.com for example.
-        my $host = $uri->host && eval { $uri->ihost };
-        if ( !$host ) {
-            $host = $url_string;
-        }
-        return $host;
+        eval {
+            my $uri = URI->new($url_string);
+            if ( !$uri->scheme ) {
```
oalders, line 96: Some if this can also be done with `URI::Heuristic` https://metacpan.org/pod/URI::Heuristic#SYNOPSIS

lib/MetaCPAN/Web/Controller/Source.pm
```diff
@@ -26,6 +26,7 @@ sub index : Path : Args {
             $c->model('API::Module')->get(@module),
         );
     }
+    die "Bad module supplied\n" if grep $_->{code} > 399, $source, $module;
```
oalders, line 29: Should we capture the module name and response code here or is that happening somewhere else?

mohawk2: No, it won't be captured anywhere. My thinking is the user will see the message, and act accordingly?

oalders:
Thanks @mohawk2!

log4perl_prod.conf
```diff
@@ -4,14 +4,30 @@ log4perl.filter.GazelleWarnings                 = Log::Log4perl::Filter::StringM
 log4perl.filter.GazelleWarnings.StringToMatch   = failed to writev: -1 errno:32
 log4perl.filter.GazelleWarnings.AcceptOnMatch   = false
 
-log4perl.appender.OUTPUT                            = Log::Log4perl::Appender::Screen
-log4perl.appender.OUTPUT.stderr                     = 1
-log4perl.appender.OUTPUT.Filter                     = GazelleWarnings
-log4perl.appender.OUTPUT.layout                     = PatternLayout
-log4perl.appender.OUTPUT.layout.ConversionPattern   = [%d] [%p] [%X{url}] %m%n
+log4perl.filter.NetworkErrors                   = Log::Log4perl::Filter::StringMatch
+log4perl.filter.NetworkErrors.StringToMatch     = Timed out at
```
haarg, line 8: Do we really want to filter out messages like this? If we're getting timeouts, we should be seeing those errors.

PR #2355, opened 2020-07-16
update bootstrap to v3.4.1

haarg:


oalders:
There are at least a couple of security warnings on this. Do they affect us?

haarg:
I accidentally picked the wrong release to use for this.  Updated to use the latest Bootstrap 3 release.

The security issues mentioned don't impact us. We control all of the HTML that could trigger the issues.

PR #2374, opened 2020-08-13
use correct path for non-latest pod pages

haarg:
The value in `@path` may not be a file path, so we should use the path
from the file we're querying instead.


heartmor:
Duplicate of #

PR #2383, opened 2020-10-01
Update recent-bar.html

matt-nelson1:


karenetheridge:
What are these sites and how are they affiliated with metacpan?

oalders:
This is actually our newest sponsor coming on board. :)

Issue #1432, opened 2014-12-29
Feature Request: Alternative Colour Schemes for syntax highlighting


kentfredric:
Because people tend not to really agree on how syntax should be highlighted, and there's also potential problems in perception[1], it would be "nice" to have an option to alternate between different syntax highlighting themes. 

Search.cpan.org has a similar feature, although its rarely used or even known about because the UI hides it _right_ at the bottom of the page.

http://search.cpan.org/~ether/Moose-2.1403/lib/Moose/Manual/Concepts.pod#COPYRIGHT_AND_LICENSE

![screenshot - 123014 - 01 29 52](https://cloud.githubusercontent.com/assets/44790/5568550/61e220f8-8fc3-11e4-8a52-2cffe9810e05.png)

My initial suggestion would be to render a very transparent gear of some kind in the upper right corner of a code section, becoming opaque when hovered over, which gave a menu of choices with regards to syntax highlighting.

Granted a determined user could simply inject a custom user stylesheet, but that's rather non-trivial and challenging from a users perspective.

<sub>1: ie: not everyone perceives colour the same, and not all displays render colour contrast the same way, have you ever had 2 different screens from different manufacturers side by side, ever tried getting them to look the same? Ugh. And its not just a case of "Make more contrast" to fight low contrast display issues, sometimes you want _less_ contrast because it could be too bright for you or something like that.</sub>

PR #2386, opened 2020-10-03
Closes metacpan-web#2182

ThisUsedToBeAnEmail:
Implementation of a OTT dark mode

- Extends user accounts with a new theme endpoint/page
- Extends activity to accept a passed in color
- Adds MetaTheme.js which is the main JS file for managing themes
- Adds a color picker https://github.com/PitPik/colorPicker
- Adds a new dependancy Colouring::In


ThisUsedToBeAnEmail:
This also resolves - (https://github.com/metacpan/metacpan-web/issues/1432)

atoomic:
Thanks for all this work, cannot wait to start using it!

oalders:
Thanks for this. @haarg will be handling the code review on this PR, so we will wait for him to give his feedback.

cpanfile.snapshot
```diff
@@ -2048,11 +2060,6 @@ DISTRIBUTIONS
       HTTP::Date 0
       Module::Build::Tiny 0.035
       perl 5.008001
-  HTTP-Lite-2.44
-    pathname: N/NE/NEILB/HTTP-Lite-2.44.tar.gz
-    provides:
-      HTTP::Lite 2.44
-    requirements:
```
oalders, line 2055: This is causing the Dockerless failure.

oalders (changes requested):

cpanfile.snapshot
ThisUsedToBeAnEmail: I'm not entirely sure how to fix this, other than manually adding back that dependency?

oalders: One really easy way to do this would be to blow away that commit and then go to the last passing Github action run for "Dockerless MetaCPAN". That has a build artifact which should include a `cpanfile.snapshot`. Overwrite your current snapshot with that one, commit it and you should be all set, since we know the tests pass with that generated snapshot. In this case, this should be the link to that build artifact: https://github.com/metacpan/metacpan-web/suites/1289931044/artifacts/20278695

ThisUsedToBeAnEmail: That seems to have worked, thanks for the guidance.

ThisUsedToBeAnEmail:
If interested I've ported the code into a chrome extension, while it waits for approval in the web store, you can install by cloning https://github.com/ThisUsedToBeAnEmail/metacpan-theme and uploading into chrome://extensions.

oalders:
I saw your demo yesterday and it does look really slick. I was wondering if you'd be willing to take care of the merge conflicts and we could put this up on a staging server to play around with?

My biggest concern here is the size of the diff and that we'd be on the hook to fix any issues in future. Maybe we can look at seeing how we reduce the size of the code changes? It looks like some of the repetitive template bits could be converted to loops etc.

Also, is it a big change to use LocalStorage to start with? That would allow a staging version to work without requiring an API deployment.

ThisUsedToBeAnEmail:
I actually think the extension is a better implementation than this merge request, so I'm not sure it should be merged. The main feature I like about the extension is that you can change the colours/scheme while viewing the different pages of the site. Something you can't currently do with the native version, and this makes it much simpler to create your own theme. Also if you look at the uptake of the extension (which is really low), I'm not sure the demand is there.

I think a more sensible approach for a native implementation is just the dark mode theme styles with a switch to enable.

oalders:
That's fair. I will close this then. Thanks for all of your efforts on this!

PR #2389, opened 2020-10-09
Maybe purge Fastly cache after author profile updated

### LINKED ISSUE #2379: Clear author page from cache if author updates their profile on MetaCPAN
If you update your profile in MetaCPAN, for example to change your email address, then your author page won't reflect the change until the cache is cleared.

This really @oalders's ticket, as he suggested that maybe you should do the equivalent of `curl -XPURGE https://metacpan.org/author/NEILB` when someone updates their profile.

oalders:
Fixes #2379


haarg (approved):

oalders:
Thanks @haarg!

Issue #1786, opened 2016-10-12
Bad layout when printing


haarg:
When trying to print pod pages, the layout is rather poor.

 - The TOC should be laid out as normal text rather than in a possibly right aligned box.
 - Most of the sidebar content could be eliminated.
 - Both sidebars distract from the main content.  Moving them after the main content would probably be preferred.
 - The search bar could be eliminated.

skrp:
I am forced to use Windows at work. 
As haarg suggested, even with the content bar hidden there are issues.
The left bar: From Date -S.C.O. section. This overflows into the contents on Windows 7-10 regardless of browsers/versions

PR #2390, opened 2020-10-09
enhance POD layout when printing

ThisUsedToBeAnEmail:


root/wrapper.html
```diff
@@ -163,7 +163,7 @@
             </div>
         </nav>
 
-        <div class="container-fluid">
+        <div class="container-fluid" id="page-<% c.action.namespace %>">
```
haarg, line 166: We try to avoid unneeded ids so they don't conflict with pod content.

root/static/less/print.less
```diff
@@ -1,4 +1,44 @@
...
     .pod { margin: 0px; border: 0px solid #000; padding: 0px}
     .home, .header, .search-bar, .box-right, .footer, a.favorite {display:none}
+    #page-pod {
+        ul.nav-list.slidepanel.sticky-panel-top,
+        ul.nav-list.slidepanel.sticky-panel-bottom.sticky-panel-sticky,
+        ul.nav-list.box-right.hidden-phone.dependencies,
+        form.search-form.form-horizontal,
+        .head-small-logo.col-md-3,
+        span.dropdown,
+        div#libwww-perl-fav,
```
haarg, line 11: Some of these selectors seem a bit weird.

haarg:
Addresses #1786

root/wrapper.html
ThisUsedToBeAnEmail: This allows for page specific CSS without including custom stylesheets.

ThisUsedToBeAnEmail: Perhaps I shouldn't be using c.action.namespace to achieve this.

haarg: classes should be preferred to ids.

`c.action.namespace` is probably the wrong thing.  I'd say anything accessing `c` is probably wrong. Not sure what the best option is though. Possibly something injected by the view class.

root/static/less/print.less
```diff
@@ -1,4 +1,45 @@
...
     .pod { margin: 0px; border: 0px solid #000; padding: 0px}
     .home, .header, .search-bar, .box-right, .footer, a.favorite {display:none}
+    #page-pod {
+        .sticky-panel-top,
+        .sticky-panel-sticky,
+        .dependencies,
+        .search-form,
+        .head-small-logo,
+        .dropdown,
+        #libwww-perl-fav,
```
haarg, line 11: This selector still doesn't make any sense.

root/wrapper.html
ThisUsedToBeAnEmail: I've changed the ID to be a class, however I'm unsure whether you will like the new approach.

root/static/less/print.less
ThisUsedToBeAnEmail: My bad, I've changed that ID to a class .logged_in

oalders:
Thanks @ThisUsedToBeAnEmail! I've merged the commits at the command line.

PR #2402, opened 2020-10-28
fix condition for showing pod link on examples

### LINKED ISSUE #2401: Examples always show "[pod]" links (which doesn't link to the pod!), even if the example does not include any pod.
See screenshot:

Screenshot from 2020-10-28 16-05-01

Each example includes a "[pod]" link as you can see. However, only some of the examples in my distribution have pod included, so for the ones that do not, there shouldn't be a link. Clicking on the "[pod]" links takes you to the example file's source code, exactly the same as clicking the filename.

A few months ago, I believe this worked correctly.

haarg:
Fixes #2401


oalders (approved):
👍

PR #2414, opened 2020-12-23
Add attributes to search input to disable autocorrect, autocapitalize, and spellcheck

mannih:
Using metacpan on mobile, my phone likes to drive me nuts by auto-correcting my
search terms, uppercasing stuff and finally putting red squiggles under the
search term. I usually search for package names and unfortunately their names
often aren't in my phone's dictionary. Those three attributes should keep all
mobile browsers from meddling with search terms.

### DIFF
--- a/root/home.html
+++ b/root/home.html
@@ -7,7 +7,7 @@
   <form action="/search">
     <input type="hidden" name="size" id="search-size" value="20">
     <div class="form-group">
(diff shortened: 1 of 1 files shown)


mohawk2:
Thank you! Should be visible in an hour or so.

mannih:
And it is working!
Thank you for all your good work on metacpan!
Happy holidays!

oalders:
Thanks for confirming that it works @mannih  and belated Happy Holidays to you as well. :)

PR #2418, opened 2021-01-05
TPF support for new developers survey

andrewsolomon:
This link needs to be removed at the end of January 2021


oalders (changes requested):

oalders (approved):

andrewsolomon:
Is it ok if we leave this in until February 15? I figured it might be sensible to leave the survey open in case the FOSDEM talk generates more opinions...

oalders:
That sounds fine!

PR #2422, opened 2021-01-20
add rt.cpan sunset notice

MartinMcGrath:
No end date specified in the comments due to the current situation. Worth putting front and centre as this has not been well communicated as far as I can tell.

References:

https://perl.topicbox.com/groups/infrastructure/Te92859202f730470

https://www.nntp.perl.org/group/perl.perl5.porters/2021/01/msg258825.html

PR #2423, opened 2021-01-21
home.html - render alerts as rows

MartinMcGrath:
[As requested](https://github.com/metacpan/metacpan-web/pull/2422), a fix for the commit yesterday. Each alert is now an individual row, and renders as requested on mobile and large screens. Screenshots attached.

Screenshot_2021-01-21_20-35-51_phone
Screenshot_2021-01-21_20-36-32_desktop


oalders:
Thanks @MartinMcGrath!

PR #2430, opened 2021-02-15
TPF survey - remove

andrewsolomon:
@oalders @haarg Thanks for having the survey here - it can now be removed.


oalders (approved):
Thanks @andrewsolomon!

Issue #2431, opened 2021-02-18
Explain in the FAQ what "Take me over?" means and how to get rid of it


szabgab:
When someone visits a CPAN module that is marked by ADOPTME or HANDOFF or NEEDHELP  they will see a banner with the tittle "Take me over?".

As a person who does not know much about Perl and CPAN I'd probably want to see some explanation what does that mean.

As a person who has experience in Perl and is interested to take over the distribution, but is not experienced in the working in PAUSE I might need a lot more explanation than the link to emails.

As a person who has just adopted a module I'd like to know how to remove that banner.

It would be nice to add an  explanation of  ADOPTME,  HANDOFF, and NEEDHELP  to the FAQ and link to it from the banner.

See also #1643

bigpresh:
Looks good, this should help clarify it.  I also think maybe the notification banner on the module page should have a brief explanation (possibly just as a little "i" info icon that links to it, so it doesn't get too wordy) - possibly just linking to the new content in the FAQ here.

oalders:
Closed via #2432

PR #2432, opened 2021-02-18
Add FAQ entry to explain module adoptions #2431

szabgab:


root/about/faq.html
```diff
@@ -151,6 +151,42 @@
...
+
+### How to become the maintainer or co-maintainer of a CPAN distribution?
+
+* TBD.
+
+### How to mark a distribution that is up for adoption?
+
+If you are the maintainer of the distribution you can follow these steps:
+
+* Visit <a href="https://pause.perl.org/">PAUSE</a>.
```
haarg, line 171: Probably better to use markdown linking here.
```
* Visit [PAUSE](https://pause.perl.org/).
```

root/about/faq.html
```diff
@@ -151,6 +151,42 @@
...
+* Check the box next to the name of the distribution.
+* Type in HANDOFF as userid.
+* Click on "Make Co-Maintainer"
+* TBD: How long till the change shows up on MetaCPAN?
+
+### How to remove the "adopt me" banner from the module?
+
+Remove the author ADOPTME, HANDOFF, or NEEDHELP from the project.
+
+* Visit <a href="https://pause.perl.org/">PAUSE</a>.
```
haarg, line 183: Same as above.

root/about/faq.html
```diff
@@ -151,6 +151,42 @@
...
+### What does the "Take me over?", "Why not adopt me?", "Looking for help!" mean?
+
+CPAN Distributions can be marked by their authors, or by the CPAN admins, in case the author is not
+available to be "up for adoption". This can be done by adding either one of the following PAUSE accounts
+as co-maintainers: ADOPTME, HANDOFF, or NEEDHELP.
+
+
+### How to become the maintainer or co-maintainer of a CPAN distribution?
+
+* TBD.
```
oalders, line 165: This section could include a link to https://neilb.org/adoption/

oalders:
@neilb do you have any blog posts that would be relevant to link to here?

root/about/faq.html
```diff
@@ -151,6 +151,42 @@
...
+
+If you are the maintainer of the distribution you can follow these steps:
+
+* Visit <a href="https://pause.perl.org/">PAUSE</a>.
+* Click on "Add Comaintainers"
+* Filter for the name of the distribution.
+* Check the box next to the name of the distribution.
+* Type in HANDOFF as userid.
+* Click on "Make Co-Maintainer"
+* TBD: How long till the change shows up on MetaCPAN?
```
oalders, line 179: This runs on an hourly cron.

root/about/faq.html
```diff
@@ -151,6 +151,42 @@
...
+### How to remove the "adopt me" banner from the module?
+
+Remove the author ADOPTME, HANDOFF, or NEEDHELP from the project.
+
+* Visit <a href="https://pause.perl.org/">PAUSE</a>.
+* Click on "Remove Comaintainers".
+* Filter for the name of the distribution.
+* Check which one of the above 3 are listed.
+* Select them and click on "Remove".
+* TBD: How long till the change shows up on MetaCPAN?
```
oalders, line 190: This runs on an hourly cron.

root/about/faq.html
oalders: I think the "TBD" is the only thing left to clarify here.

szabgab: Is there a blog post explaining I could link to?

oalders: How about http://neilb.org/2013/07/24/adopt-a-module.html ?

neilb:
Could also link to http://neilb.org/2014/01/31/adoption-request.html which is a template mail message if you want to email someone asking to adopt their module, but the wording of the template suggests a bunch of things you should have done before you're emailing someone asking to take over their dist.

oalders (approved):
Thanks for this @szabgab! Let's go live with it and see what feedback we get.

szabgab:
Thank you. When is this expected to be deployed?

oalders:
This will have already been deployed. Deployments happen every 30 or 60 minutes. (Can't remember which).

You can purge the cache for any individual page via 

```
curl -XPURGE https://metacpan.org/path/to/page
```

If you can't see your changes immediately after a purge, please let me know.

szabgab:
Thanks. That worked and I can see the new content, but I don't understand why do we need to do this manually?

oalders:
The caching is done via Fastly and our deployments are not sophisticated enough to pick up all the correct URLs to purge on arbitrary code merges. We do catch some cases, but not all of them. They'll eventually expire, so it's generally not a huge issue, but we occasionally purge some URLs manually or just purge everything.

PR #2434, opened 2021-02-20
rt not closing banner on home page

szabgab:


oalders (approved):
Thanks @szabgab!

MartinMcGrath:
Would it make more sense for this to use the `alert-info` class rather than alert-`danger`?

szabgab:
Either way IMHO it should be removed within a few days, maybe a week.

Issue #1435, opened 2015-01-07
Visually distinguish deprecated modules based on x_deprecated field


haarg:
In #toolchain, there has been discussion of establishing x_deprecated as a field to indicate deprecated modules.  In the future, this could be used in search ranking.  Initially though, it would be good to add a visual indicator of this somewhere when viewing modules.

oalders:
+1

oalders:
I saw some discussion about this being a boolean field for the dist.  Is there any plan for deprecating just one module within a distribution or is that not a common enough use case?

haarg:
@dagolden suggested handling per-module deprecations via x_deprecated fields in the provides hash.  I think this is a sensible solution.

oalders:
Once people start using this, could someone update this ticket with a couple of example dists that we can work from?

oalders:
https://metacpan.org/release/POEx-Role-PSGIServer

karenetheridge:
http://grep.cpan.me/?q=x_deprecated

https://metacpan.org/release/ETHER/Directory-Scratch-0.18
https://metacpan.org/release/ETHER/Any-Moose-0.26
https://metacpan.org/release/JLUIS/MetaCPAN-API-Tiny-1.150270
https://metacpan.org/release/BARBIE/Test-YAML-Meta-0.22

karenetheridge:
>  Is there any plan for deprecating just one module within a distribution or is that not a common enough use case?

Yes, I've got this feature written for Dist::Zilla::Plugin::Deprecated, and will release it as soon as the necessary change to CPAN::Meta::Merge is released (likely at/by Berlin).

karenetheridge:
@oalders these distributions now have individual modules marked as deprecated (vs. the entire distribution):

https://metacpan.org/release/Dist-Zilla-Plugin-Test-EOL

oalders:
@karenetheridge I don't see x_deprecated in https://metacpan.org/source/ETHER/Dist-Zilla-Plugin-Test-EOL-0.17/META.json  Am I looking at the wrong release?

karenetheridge:
yes, it's 0.18, just uploaded. :)

oalders:
Ah. I see it in https://metacpan.org/source/ETHER/Dist-Zilla-Plugin-Test-EOL-0.18/META.json  Thanks!

mickeyn:
now that we have the field in the metadata, this can be done.

bluefeet:
[...]

PR #2436, opened 2021-02-21
Show notifications for Deprecated modules/distributions

bleargh45:
Using similar visual indicators to what we have in place already for ADOPTME, NEEDHELP, and HANDOFF, provide an indication that a Module has been Deprecated, or possibly the entire Release/Distribution.

Relates to metacpan/metacpan-web#1435

### DIFF
--- a/root/inc/notification.html
+++ b/root/inc/notification.html
@@ -24,6 +24,10 @@
             text   = "The maintainer of this distribution is looking for someone to take over!",
             body   = "Hello,\r\n\r\nI'm interested in taking over the %s module. If you can grant me the privilege I will release the next version.\r\n\r\n"
         },
+        DEPRECATED = {
+            title  = "Deprecated.",
+            text   = "The maintainer of this distribution has indicated that it is deprecated and no longer suitable for use.",
+        },
     }.$notification_type;
     NOTIFICATION.mailto = MAILTO.sprintf(
         notification.emails.join(','),
--- a/root/inc/notification.html
+++ b/root/inc/notification.html
@@ -40,10 +40,12 @@
         <h2><% NOTIFICATION.title %></h2>
         <span>
             <% NOTIFICATION.text %>
-            If you're interested then please contact
-            <% NOTIFICATION.contactwhom %> via
-            <a href="<% NOTIFICATION.mailto %>">email</a>.
-            <a href="/about/faq#howtoadoptadistribution"><i class="fa fa-info-circle"></i></a>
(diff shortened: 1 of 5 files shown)


root/inc/notification.html
```diff
@@ -24,6 +24,17 @@
             text   = "The maintainer of this distribution is looking for someone to take over!",
             body   = "Hello,\r\n\r\nI'm interested in taking over the %s module. If you can grant me the privilege I will release the next version.\r\n\r\n"
         },
+        DEPRECATED = {
```
haarg, line 27: The content of this section looks wrong.

root/inc/notification.html
```diff
@@ -24,6 +24,17 @@
...
+            title  = "Deprecated.",
+            subject = "Take over %s?",
+            contactwhom = "them",
+            text   = "The maintainer of this distribution has indicated that it is deprecated and no longer suitable for use.",
+            body   = "Hello,\r\n\r\nI'm interested in taking over the %s module. If you can grant me the privilege I will release the next version.\r\n\r\n"
```
Grinnz, line 32: Deprecating a module isn't really an indication of looking for maintainers by itself. There isn't any reason to present a mail option.

bleargh45: Marvelous!  That could actually help make this simpler...

By not having to provide an indication that we could be looking for someone to take over the module, the implementation for this could be refactored and simplified further.

root/inc/notification.html
bleargh45: Wrong in what sense?  Similar to what @Grinnz had commented on, in that we do not need to provide an indication that the author may be looking for someone to take over the distribution?

root/inc/notification.html
bleargh45: Shall I proceed ahead with some refactoring here, in the mind of setting this up such that when a Distribution is marked as Deprecated, that we do not provide the statement/option to e-mail the author?  After which point, I can then amend the PR w/new commits.

oalders: Sounds good!

bleargh45:
Ok... second attempt.  :)

bleargh45:
And... think I've figured out why those tests are failing... there's an extra `"` in the HTML being used by the Notifications right now, so let me yank that out and try this again.

bleargh45:
Yup, there it is... that should fix the running tests... fix the HTML first.

oalders (approved):
Thanks @bleargh45!

Issue #2438, opened 2021-02-22
Deprecation banner on module which doesn't appear to be deprecated


bigpreshkt:
I see that #2436 has started showing deprecation banners.

I see one for, e.g. https://metacpan.org/pod/CHI

However, I don't see anything to suggest that module is deprecated, and if I ask the API:

```
[davidp@columbia:~]$ jq .deprecated <(curl -s https://fastapi.metacpan.org/v1/release/CHI)
"false"
```

Similarly, no sign that it should be deprecated at https://explorer.metacpan.org/?url=/module/JSWARTZ/CHI-0.60/lib%2FCHI.pm

Am I missing something obvious, or is this displaying the banner wrongly?

bleargh45:
That _does_ appear to be incorrect; let me go take a look at that for you right now.

bleargh45:
Now seeing what the discrepancy was that caused this to occur...

When using the local `metacpan-docker` setup to do development with, the `deprecated` field is properly cast as a Boolean, and returns `false`.  The Production API, however, returns it as a string value, `"false"`:

```
$ curl --silent http://localhost:5000/module/CHI        | jq .deprecated
false

$ curl --silent http://fastapi.metacpan.org/module/CHI  | jq .deprecated
"false"
```

bleargh45:
And, more frustratingly, it also appears that the boolean-ness of the values are not consistent across the API Index:

```
$ curl --silent http://fastapi.metacpan.org/release/Alien-Prototype-Carousel  | jq .deprecated
true

$ curl --silent http://fastapi.metacpan.org/release/CHI  | jq .deprecated
"false"
```

```
$ curl --silent http://fastapi.metacpan.org/module/Dist::Zilla::Plugin::EOLTests | jq .deprecated
true

$ curl --silent http://fastapi.metacpan.org/module/Dist::Zilla::Plugin::Test::EOL | jq .deprecated
"false"
```

So... for the moment, simplest approach looks like checking not just for the 1/0-ness of the Bool, but also for "true/false" (as strings).  At least, until we can determine what's up with the underlying index, and can figure out how to address that.

oalders:
@mickeyn any idea what's going on with the booleans?

bleargh45:
[...]

PR #2439, opened 2021-02-22
Disable "deprecated" notifications

bleargh45:
While this worked great in Dev on a partial index, something fishy is going on and the notification is showing incorrectly in Production.

Until we can diagnose the cause and correct, disable the display of Deprecated notifications.

Related to #2438


ssoriche (approved):
👍

oalders:
Maybe we could add a regression test with the next iteration?

bleargh45:
Appreciate the quick action to get that merged/deployed!

And yes... next up is a regression test to try to verify all of the various cases of "distribution/module deprecated (or not)".  I will also see what I can do to get my local `metacpan_web` container connected to the real/production API, so that I can attempt to reproduce the problem that way.

PR #2441, opened 2021-02-28
Re-enable "deprecated" notifications

bleargh45:
This reverts commit 19df9e5be15a14ef4f8fa44d15eca7c9d8046ae2.

Re-enables "Deprecated" notifications.

As per discussion in metacpan/metacpan-web#2438, the data in ElasticSearch has been updated and now returns proper boolean values (instead of potential string `"true"` or `"false"`).

With the data corrected, we should now be able to re-enable these notifications for deprecated Modules/Distributions.


oalders (approved):
Thanks! Let's see what happens. :) 👍

PR #2449, opened 2021-03-27
sidebar width become 174px to 195px

yuki-kimoto:


mohawk2:
I think #2450 is a better approach.

haarg:
I've merged #2450, so this is no longer needed.

oalders:
Thanks for all of your help with this, @yuki-kimoto!

PR #2450, opened 2021-03-28
fix sidebar width

haarg:
This fixes the sidebar width to match the new font size. It adjusts various css rules to be calculated based on the sidebar width, rather than hard coded. It also changes the sidebar width to be calculated based on the font size.

Replaces #2449


mohawk2 (dismissed):

mohawk2:
Looks good! On 5.30 it's failing `t/controller/changes.t` with:

```
#   Failed test 'link to suggested release'
#   at t/controller/changes.t line 44.
#                   ''
#     doesn't match '(?^:/test-dist-name-that-does-not-exist-i-hope$)'
```

haarg:
That error is due to the new release of Template::Alloy and not related to this PR. master is now failing its tests as well for the same reason.

PR #2452, opened 2021-04-12
remove mirrors page

haarg:
The CPAN mirror network has been retired, and now only www.cpan.org is a
listed mirror.


oalders (approved):
🌐

PR #2455, opened 2021-04-15
Add @daxim's user styles to author

mjgardner:
See https://www.reddit.com/r/perl/comments/mpbgsd/readable_distro_names_and_dates_on_metacpan/

### DIFF
--- a/root/static/less/author.less
+++ b/root/static/less/author.less
@@ -92,3 +92,11 @@
         margin-right: -15px;
     }
 }
+
+table#author_releases {
+    table-layout: inherit;
+}
+
+a.ellipsis span {
+    max-width: inherit;
+}


haarg:
This doesn't do what it's intended to do without the `!important`, which overrides the other styles. Doing this correctly would involve removing the abbreviation code.

All this does is remove all width limits on the release table. This pushes the dates off the right side of the screen if the dist names are longer. It also makes the content overlap the sidebar if you scroll down and to the right.

haarg:
The screenshots in the thread are also not very helpful in determining the intent of these changes, because they don't include what browser and OS is being used. None of the browsers I have access to will render like the 'before' screenshot.


There is evidence that daxim has applied other user styles, which makes it hard to tell what is due to their browser, and what they have changed themselves.

wchristian:
after trying to replicate his screenshot i suspect this might have to do with device resolution dependence. on my really big monitor i have to zoom in ALL the way AND narrow the window to trigger name abbreviation

mjgardner:
I suppose you can ignore this PR. I mainly did it to show him that it's not hard to contribute to this project. But he doesn't want to be shown.

oalders:
Thanks for your efforts @mjgardner!

PR #2463, opened 2021-04-20
update monospace fonts

haarg:
DejaVu Sans Mono is not the best looking font, and also has some issues:
https://gitlab.gnome.org/GNOME/pango/-/issues/392

Replace it with Liberation Mono, which is also included on many debian
systems. Also re-order the fonts to prefer Consolas on most Windows
systems.


oalders (approved):
👍

PR #2464, opened 2021-04-21
clean up diff model to provide more reasonable data to template

haarg:
The data from the API for diffs is not in the best form. Eventually that
should get fixed, but for now move some logic into the model layer
rather than the template.


oalders (approved):
👍

PR #2468, opened 2021-04-26
microdata schema updates

haarg:
Removes the SoftwareApplication microdata, since it seems misused and isn't accepted by Google anyway in its current form.

Adds WebSite data to the front page to allow google to show a search box for the site.


oalders:
Looks like it's failing on:

```
#   Failed test 'no inline script in root/home.html'
#   at t/html.t line 22.
```

oalders (approved):
Thanks!

PR #2470, opened 2021-04-28
be explicit about perl critic policies

haarg:
Rather than relying on the theme and severity which are unpredictable,
explicitly list the perl critic policies we want to use.

Also remove all the excluded files from the critic/tidy checks, since
they are all passing now.


oalders (approved):
Looks great!

PR #2471, opened 2021-04-28
use divs to lay out plusser lines rather than p

haarg:
They aren't really paragraphs, so use div tags to have more natural
markup. Apply margin using CSS rather than relying on <p> tag.

Also remove trailing period from non-pause user count to be consistent with
pause user count.


oalders (approved):
👍

PR #2475, opened 2021-04-28
always provide distnameinfo with release data, and use it in templates

haarg:
Rather than getting specifically the version from CPAN::DistnameInfo in
one case, provide it for all release data. This is then used to simplify
part of the template that was re-calculating the version from the dist
name.


oalders (approved):
👍

PR #2476, opened 2021-04-29
refactor login handling

haarg:
Previously, logins were handled by embedding a lot into the templates,
requiring special handling for OpenID and PAUSE. It also required the
oauth urls and keys to be available on every page, which meant it had to
have special handling in the core view module.

Move to using separate end points for each login type, which then can do
their own handling as appropriate. For OpenID and PAUSE, this means
showing a form. For others, it's an immediate redirect. Now only the
login controller needs to know anything about the oauth keys and URLs.
The templates can be simpler because they don't need any per-account
handling. And it adds a form for the PAUSE login, rather than the ugly
javascript prompt.


oalders (approved):
👍

mohawk2:
Superb!

PR #2480, opened 2021-04-29
improve markup of search page

haarg:
Use more semantic tags and CSS for search page layout. The result is a
couple pixels different, but it's not a meaningful difference.


oalders (approved):
👍

PR #2490, opened 2021-05-05
convert templates to Text::Xslate

haarg:
Text::Xslate is meant to be faster than other templating engines, and
provides some other capabilities to make templates more reusable, such
as template inheritance. Previously we were using Template::Alloy, but
it has some inconsistent behaviors, such as either not outputting
overloaded objects, or not applying automatic escaping to them,
depending on which version is used.


oalders (approved):
Thanks, @haarg! 💯

PR #2491, opened 2021-05-05
use xpath for all tests

haarg:
We were only using CSS selectors for one test, which can trivially be
switched to xpath. This makes the tests more consistent, removes an odd
interface, and removes a prereq.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -36,7 +36,6 @@ requires 'Gravatar::URL';
 requires 'Hash::AsObject';
 requires 'HTML::Escape';
 requires 'HTML::Restrict', '2.2.2';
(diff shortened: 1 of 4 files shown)


oalders:
Looks like a tidying error and I'm also seeing:

```
XPath error : Invalid predicate
//div[contains-token(@class, "content")//pre[@id=source]
                                                        ^ at /__w/metacpan-web/metacpan-web/local/lib/perl5/Test/XPath.pm line 103.                                                        ^ at /__w/metacpan-web/metacpan-web/local/lib/perl5/Test/XPath.pm line 103.
```

oalders (approved):
Looks great! 🧹

PR #2495, opened 2021-05-10
use GitHub as author for snapshot update PR

haarg:


oalders:
This will prevent @ssoriche from gaming the leaderboard.

PR #2498, opened 2021-05-11
Refactor authentication

haarg:
This refactors the authentication to better fit in to the Catalyst Authentication plugin. It moves the login handling out of the controller, and instead into the auth system and user model.

The custom Realm stores the auth token in the session in the way we want it, rather than that having to be done externally.

The custom Store replaces the Proxy store. The Proxy store creates User objects at the wrong time. Replacing it means the user methods it uses can be class methods, as they should be.

The user object can now behave normally, being initialized with relevant data. It doesn't need to dispatch to a separate object, but can instead have whatever methods and attributes are useful. Instead of needing to carry around the access token externally and pass it to every relevant API call, the user object itself stores the token and knows to pass it along when performing those actions.


oalders (approved):
👍 🔑

Issue #1618, opened 2015-11-12
[minor] Hidden part of Changes should not be too small.


van-de-bugger:
meta::cpan hides a part Changes if Changes too long and shows `Show More` link to show full text. Generally it is ok, but hidden part should not be too small. For example, there is no pint in hiding one line because `Show More` occupies all saved space. I would say hidden part should be longer than 5 lines of text.

See the [example](https://metacpan.org/release/VDB/Systemd-Daemon-0.07). Part of Changes is hidden. Click on `Show More` shows *one empty line*. It's ridiculous.

rwstauner:
Patches welcome.

oalders:
Closed via #2502

PR #2502, opened 2021-05-14
fix changes display collapsing

haarg:
Previously it was possible for the show/more less button on the changes
block to be displayed even if the hidden content was a single line,
which was rather pointless. And after changing the font size, the
calculated sizes could be off by a fraction of a pixel, resulting in
showing the button but it doing nothing.

Fix collapsing to not apply if javascript is disabled, and include the
size of the button when calculating if it should be collapsable.

Also tweak the collapse height to be an integer multiple of the line
height, and remove some unneeded padding between change groups.

### DIFF
--- a/root/release.tx
+++ b/root/release.tx
@@ -13,7 +13,7 @@
   %%  }
 </ul>
 %%  }
-<div id="last-changes" class="well collapsed">
+<div id="last-changes" class="well">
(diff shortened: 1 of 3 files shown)


haarg:
This should fix #1618

oalders (approved):
👍

PR #2504, opened 2021-05-14
force install all transitive modules when building snapshot

haarg:
The cpanfile.snapshot file is generated using a version of perl that may
not match what we are deploying on. Additionally, some dependencies may
not predictably be installed with a normal cpm/carton run. Both of these
cases can end up with modules missing from the snapshot which are
required when installing on various systems.

We could try to provide cpm with the deployment perl version, but that
means it may not be appropriate for other perl versions.

Instead, we provide a list of modules that should always be installed,
even if they are provided by core. We use cpanm to extract these prereqs
from a secondary cpanfile, then pass them to cpm to install. cpm can't
use the alternate cpanfile directly, because it only pays attention to
the --reinstall flag when provided a module list. cpanm is given both
--installdeps and --showdeps, because only "--installdeps ." will look
at a cpanfile. The --showdeps option overrides its actual behavior,
showing the prereqs rather than installing them. Yes, this is a horrible
hack. It also needs the -L option, even though it it isn't installing
anything, to silence warnings about unwritable paths.

### DIFF
--- a/.github/workflows/update-snapshot.yml
+++ b/.github/workflows/update-snapshot.yml
@@ -8,14 +8,33 @@ jobs:
         runs-on: "ubuntu-20.04"
         steps:
             - uses: actions/checkout@v2.3.4
+            - name: Get cpm
+              run: >
+                  curl -sL -o cpm https://git.io/cpm;
+                  chmod +x cpm
             - name: Install deps
               run: >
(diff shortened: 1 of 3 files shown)


oalders (approved):
👍

PR #2506, opened 2021-05-16
fix hamburger menu not appearing

### LINKED ISSUE #2505: Missing hamburger icon?
I think metacpan used to have ["hamburger icon"](https://en.wikipedia.org/wiki/Hamburger_button) in the left top for mobile site
so that we could see some useful links such as "Browse", "Changes", etc.
But now it's gone.

Screen Shot 2021-05-16 at 20 45 08

It would be nice if metacpan has the "hamburger icon" again.

haarg:
This was broken by #2490.

Fixes #2505


oalders (approved):
LGTM

PR #2510, opened 2021-05-23
Several minor cleanups

haarg:
Expand perl tidy options to be more readable.

Return JSON from JSON end points even on errors.

Per-page classes should be part of the templates, not the controllers.

Improve a few links on the about page.

Silence a warning in the encoding test.


oalders (approved):
👍

PR #2511, opened 2021-05-23
Revert "Mark flaky Moose Pod content comparison test as TODO"

haarg:
This test should always pass, and always seems to. This generates noise
about a passing TODO test. The logs of it failing are no longer
available, making debugging difficult.

The restored check will either continue to work, or will help reveal the
failure and allow a proper fix.

This reverts commit 8821b4a8f0be06ffd332be68d48cf0c4e9e38be5.


oalders (approved):
👍

PR #2512, opened 2021-05-23
Refactor URLs for more consistency

haarg:
This refactors the URLs of the site to be based on the object is is related to, rather than the operation it is. So rather than:

/permission/author/AUTHOR
/permission/distribution/My-Dist
/changes/distribution/My-Dist
/changes/release/AUTHOR/My-Dist-0.01
/contributing-to/My-Dist
/contributing-to/AUTHOR/My-Dist-0.01

The new URLs look like 

/author/AUTHOR/permissions
/dist/My-Dist/permissions
/dist/My-Dist/changes
/dist/My-Dist/contribute
/release/AUTHOR/My-Dist-0.01/changes
/release/AUTHOR/My-Dist-0.01/contributing

This provides better consistency on how to construct future URLs.

The new roots are `/author/$AUTHOR`, `/dist/$dist`, `/release/$AUTHOR/$release`, and `/module/$module`.

`/dist/$dist` serves as the home both for things that are related to the distribution itself, and the latest release. Generally, anything available under `/release/$AUTHOR/$release` is also available under `/dist/$dist`.

This chaining also provides a common place to attach behavior, such as redirecting to upper case authors, instead of needing to replicate that in many actions.

Pod viewing URLs previously all started with `/pod/`. Rather than retaining this naming, the new URLs use `*/view/`. This will allow it to be used for viewing other file types, such as markdown, which currently is done by taking over the source page. The `/pod/` root is maintained for modules, because they will always be using pod.

### DIFF
--- a/app.psgi
+++ b/app.psgi
@@ -133,6 +133,10 @@ builder {
         );
     }
 
+    enable 'FixMissingBodyInRedirect';
+
+    enable '+MetaCPAN::Middleware::OldUrls';
+
     enable '+MetaCPAN::Middleware::Static' => (
         root     => $root_dir,
         dev_mode => $dev_mode,
--- a/cpanfile
+++ b/cpanfile
@@ -67,6 +67,7 @@ requires 'Plack::Middleware::Session::Cookie';
 requires 'Plack::Session';
 requires 'Plack::Test';
 requires 'Ref::Util', '>= 0.008';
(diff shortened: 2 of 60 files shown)


oalders (approved):
Looks great!

oalders (approved):
Thanks, @haarg!

PR #2517, opened 2021-05-24
always use solid icons

haarg:
FontAwesome has solid and outline variants for most of its icons. We are
using the solid variant for nearly every icon.

Change the remaining few icons from outline to solid for increased
visual consistency.


oalders (approved):
👍

PR #2520, opened 2021-05-31
Add instagram to profile options

benkasminbullock:


haarg (changes requested):
This will also need a 16x16 icon in `root/static/profile/`

oalders (dismissed):
Thanks @benkasminbullock!

benkasminbullock:
> This will also need a 16x16 icon in `root/static/profile/`

There is an SVG icon here:

https://github.com/edent/SuperTinyIcons/blob/master/images/svg/instagram.svg

Here is a 16x16 of that:

instagram

haarg (approved):

PR #2527, opened 2021-06-11
Add png icons for accounts for edit profile page

benkasminbullock:
The icons are not currently appearing in the edit profile page, I think this is what needs to be added.


benkasminbullock:
Just to be clear, here is what I am referring to:

icons

haarg (approved):

PR #2528, opened 2021-06-12
Make scrollable area larger for pod pre areas

omega:
The problem is that `.pod pre` and `.syntaxhighlighter` both apply `overflow` in some way. By turning off `overflow-x` on `.syntaxhighlighter`, we end up with the outer `.pod pre` scroll area, which is larger, since that is the one that applies `padding`. This makes it easier to scroll horizontally on one-line code blocks on mobile.

I noticed this on https://metacpan.org/dist/App-perlimports/view/script/perlimports, it has a lot of examples in the SYNOPSIS that are hard for me with my fat fingers to scroll on my mobile.

### DIFF
--- a/root/static/less/pod.less
+++ b/root/static/less/pod.less
@@ -57,6 +57,10 @@
         margin-right: 0;
         padding: 10px;
         overflow-x: auto;
+        
+        .syntaxhighlighter {
+            overflow-x: none !important;
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


haarg:
I think this would be better as a rule:
```
overflow: visible !important;
```
in the `html body .syntaxhighlighter` section in `root/static/less/syntaxhighlighter.less`.

omega:
Thanks for the feedback, I just looked for a quick fix, moved it to the better place now (and used the valid `visible` instead of the invalid `none`)

haarg (approved):

haarg:
Thanks.

Issue #1590, opened 2015-09-14
reduce conflicts between HTML ids in content vs chrome


haarg:
MetaCPAN has HTML ids that it uses in many places for javascript interactions and styling.  These ids can sometimes conflict with the generated ids in pod content.  We should try to reduce the conflict between these as much as possible.  Possibly by prefixing all of our own ids, or at least making them more unique.

As an example of this, 67782416093099323d27f92996af7deb4812ccdf added a field with id="size", meaning that you can no longer link to a section named "size" in any pod pages.

gordon-fish:
I would like to take this one on. Feel free to put me as "assignee" (assuming I'm understanding how that works correctly.)

haarg:
Fixed by #2530

PR #2530, opened 2021-06-12
Prefixed all non-POD related ids with metacpan_

gordon-fish:
Prefixed all non-POD related ids with metacpan_

Also left out #index* ids since those also come from pod2html according to discussion in the #metacpan channel on irc.perl.org

This is to address issue #1590

### DIFF
--- a/root/about/contributors.tx
+++ b/root/about/contributors.tx
@@ -23,7 +23,7 @@
       var done = 0;
       return {
         render: function() {
-          $('#author-result-loading').hide();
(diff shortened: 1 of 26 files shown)


gordon-fish:
I have made this a draft PR for now while testing is taking place.

Anyone who would like to test, please browse around https://mc.gfish.info/ and report if anything seems to be broken (in case something was missed.)

haarg (approved):

PR #2535, opened 2021-06-16
Fixed source highlighting/line number links not showing due to a miss…

gordon-fish:
…ed prefix for #source in syntaxhighlighter.js

Related to issue #1590 and pull #2530


haarg (approved):

gordon-fish:
Thank to LeoNerd, tm604, and other in L#perl for spotting this.

PR #2536, opened 2021-06-16
purge template cache before preloading

haarg:


oalders (changes requested):
> Failed test 'lib/MetaCPAN/Web/View/Xslate.pm is tidy'

oalders (approved):
👍

PR #2538, opened 2021-06-18
Add bus factor to left panel

oalders:
Screen Shot 2021-06-18 at 11 43 12 AM

### DIFF
--- a/root/base/release.tx
+++ b/root/base/release.tx
@@ -84,6 +84,13 @@
     <li>
       <a rel="noopener nofollow" href="http://cpants.cpanauthors.org/release/[% $release.author %]/[% $release.name %]"><i class="fa fa-fw fa-line-chart black"></i>Kwalitee</a>
     </li>
+    %% if defined $distribution.river.bus_factor {
+    <li>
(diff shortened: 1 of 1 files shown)


root/base/release.tx
```diff
@@ -84,6 +84,13 @@
     <li>
       <a rel="noopener nofollow" href="http://cpants.cpanauthors.org/release/[% $release.author %]/[% $release.name %]"><i class="fa fa-fw fa-line-chart black"></i>Kwalitee</a>
     </li>
+    %% if defined $distribution.river.bus_factor {
+    <li>
+      <div id="tooltip" data-toggle="tooltip" data-placement="right" title="The # people with an indexing permission on [% $release.distribution %] who have released something to CPAN in the last 2 years">
```
haarg, line 89: This id seems wrong, but it would be better to remove the id and add the class `ttip`. Code in `cpan.js` will then set up the tooltip without needing and `<script>` tags or anything page specific.

root/base/release.tx
```diff
@@ -219,3 +226,7 @@
 
 %%  include inc::module_install;
 %%  }
+
+<script>
+    $('#bus-factor').tooltip();
+</script>
```
haarg, line 232: This section can be removed.

haarg (changes requested):

haarg (approved):

PR #2541, opened 2021-06-20
update to new fontawesome icon names

haarg:
Some icon names have changed over time, or we were using aliases. Update
to the new names so we don't need the shims.


oalders (approved):
👍

PR #2542, opened 2021-06-20
left panel should be on top

haarg:
Make sure the left panel displays on top of other content when it is
positioned.


oalders (approved):
👍

PR #2556, opened 2021-07-01
Bus factor tooltip - further explanation of purpose

Grinnz:
Some further clarification of the purpose of this metric. Not sure if it's too much to put in the tooltip. Alternatively maybe the bus factor could link to a FAQ section with an explanation of it and the river stage.


Grinnz:
I opened #2557 to add such entries to the FAQ, which may be a preferable approach.

oalders (approved):
Thanks!

PR #2584, opened 2021-07-23
Add hosted badge to MetaCPAN

thibaultduponchelle:
metacpanbadges

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Badge.pm
+++ b/lib/MetaCPAN/Web/Controller/Badge.pm
@@ -8,13 +8,12 @@ use DateTime;
 
 sub dist : Chained('/dist/root') PathPart('badge.svg') Args(0) {
     my ( $self, $c ) = @_;
-    my $dist = $c->stash->{distribution_name};
 
-    $c->forward( 'badge', [ distribution => $dist ] );
+    $c->forward( 'badge' );
 }
 
 sub badge : Private {
-    my ( $self, $c, %args ) = @_;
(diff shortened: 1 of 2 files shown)


haarg:
I'm not sure I understand the purpose of this if it's just showing a static image.

thibaultduponchelle:
> I'm not sure I understand the purpose of this if it's just showing a static image.

Ideally it should include the version but it is not implemented in this pull request.

lib/MetaCPAN/Web/Controller/Badge.pm
```diff
@@ -0,0 +1,27 @@
+package MetaCPAN::Web::Controller::Badge;
+
+use Moose;
+
+BEGIN { extends 'MetaCPAN::Web::Controller' }
+
+use DateTime;
```
oalders, line 7: This doesn't appear to be used?

oalders:
Thanks for this, @thibaultduponchelle!

> Ideally it should include the version but it is not implemented in this pull request.

Did you want to implement it in a followup or were you hoping someone else would do that?

mohawk2:
There are two failing tests: one has been fixed already so you need to rebase past that, and the other needs you to run the "tidyall".

thibaultduponchelle:
> Did you want to implement it in a followup or were you hoping someone else would do that?

The answer was "I have no idea" :)

Sorry for the delayed answer, I was in vacation, if you don't mind, I will close this *incomplete* PR for now.

PR #2607, opened 2021-09-23
Add newuser blurb to landing page

PaulGWebster:
Please test this where required I really do not do frontend... I tried to generate the css changes with this command:

```
/home/paul/Work/metacpan-docker/src/metacpan-web/root/static/less: lessc style.less > global.less
```

This caused an error about .border-radius not being defined, the internet said to change this to border-radius: N, which I did but indeed not sure of the real effect of this!


haarg:
You don't need to regenerate the CSS. That is done automatically when the server is started.

Can you remove all the CSS changes aside from what is related to the front page addition?

haarg:
The easiest way to get lessc working is by running `yarn install`. Alternatively, you can install lessc globally and add an entry like:
```
lessc_command = lessc
```
to `metacpan_web_local.conf`.

PaulGWebster:
> You don't need to regenerate the CSS. That is done automatically when the server is started.
> 
> Can you remove all the CSS changes aside from what is related to the front page addition?

that was the thing ... it did not auto generate on server start which forced me to go looking how to generate it :)

PaulGWebster:
I have just pushed the css file that is required for the text only, seems all good perhaps I had it cached not sure!

root/home.tx
```diff
@@ -37,5 +37,6 @@
         <button class="btn search-btn help-btn">?</button>
     </div>
   </form>
+  <p class="helper_blurb">Just getting started with perl? Try these <a target="_blank" rel="noopener" href="https://www.perl.org/learn.html">free learning resources</a> from the Perl community</p>
```
oalders, line 40: > Just getting started with perl?

Should that be `Perl` rather than `perl`? We're referring to the language rather than the interpreter here.

PaulGWebster: Updated to Perl

oalders (approved):
Thanks @PaulGWebster!

PR #2620, opened 2021-11-15
Gravatar link should not use current profile tab

mickeyn:


root/account/profile.tx
```diff
@@ -38,7 +38,7 @@
                     <div class="col-sm-8 col-md-6">
                         <div class="form-control-static">
                             <div><img class="author-avatar" src="[% gravatar_image($author, 130) %]" alt="Author image"></div>
-                            <div><a href="https://en.gravatar.com/emails/">Change Gravatar for [% $author.pauseid | lc %]@cpan.org</a></div>
+                            <div><a target="_blank" href="https://en.gravatar.com/emails/">Change Gravatar for [% $author.pauseid | lc %]@cpan.org</a></div>
```
oalders, line 41: Since we're using `target="_blank"`, should we also add a `rel=”noopener”` here?

mickeyn: added

oalders (approved):
Thanks, @mickeyn!

PR #2627, opened 2021-12-21
Advertise TPRC 2022 on HomePage

atoomic:
This is adding a small text banner to the HomePage for TPRC 2022 https://perlconference.us/tprc-2022-hou/

attached is a snapshot of the rendering 
Screen Shot 2021-12-21 at 3 35 28 PM

Thanks for your consideration


oalders (approved):
🎉

PR #2635, opened 2022-01-11
update logos for sponsors

mbeijen:
This is a little weird, but I started looking into this because the fastly logo on my ipad on the metacpan.org website did not look pretty 🤒


oalders (approved):
Thanks, @mbeijen!

Issue #2674, opened 2022-05-02
"Github info" table still shows `git://` URLs for github


nwc10:
**Describe the bug**

The page for modules shows a "Repository (git clone)" entry, which provides a pop-up table with various git links, titled "github info". This include a link described as "Git URL" using the `git://` protocol. eg `git://github.com/domm/App-TimeTracker.git`

github no longer supports this protocol. Attempting to use *that* URL gives this error message:

```
$ git clone git://github.com/domm/App-TimeTracker.git
Cloning into 'App-TimeTracker'...
fatal: remote error:
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

**To Reproduce**

1. Go to a module page - eg https://metacpan.org/pod/App::TimeTracker
2. Hover over "Repository"
3. A table pops up - "Github info"
4. Move to "Git URL"
5. copy the link there - git://github.com/domm/App-TimeTracker.git
6. attempt to use this link in a terminal command: `git clone git://github.com/domm/App-TimeTracker.git`
7. get error from github
 
**Expected behavior**

Just remove that line from the table.
But, *just* for github? (I don't know the site source well enough to know if there are different table formats for "everything not github", or for specific other popular hosting sites such as gitlab or bitbucket)

**Desktop (please complete the following information):**

 - OS: macOS
 - Browser Firefox
 - Version 99

(you asked. But it doesn't matter here)

PR #2675, opened 2022-05-02
remove git clone URL from github popup

haarg:
GitHub no longer allows use of the git protocol. Remove it from the
popup for GitHub repositories.


haarg:
Fixes #2674.

Amusingly, `git://` link that no longer works is still provided by the GitHub API.

oalders (approved):

PR #2676, opened 2022-05-02
clean up GitHub repository URLs

haarg:
Authors will use various forms of URLs when including repository
metadata. For known git hosts (only GitHub so far) we can clean up those
URLs to be presented better in the UI.

One example of this is including just a clone URL using git://. This
isn't friendly to present in the UI to start with, since it doesn't lead
to a web page, but also GitHub no longer supports git:// cloning. There
is also little purpose of presenting a https web link with a second link
for cloning that includes .git at the end. Either one could be used for
cloning.

### DIFF
--- a/lib/MetaCPAN/Web/Model/ReleaseInfo.pm
+++ b/lib/MetaCPAN/Web/Model/ReleaseInfo.pm
@@ -151,6 +151,8 @@ sub normalize {
                 $data->{release}{release},
                 $data->{distribution}{distribution}
             ),
+            repository =>
+                $self->groom_repository( $data->{release}{release} ),
             plussers => $data->{plussers}{plussers},
             (
                 $self->full_details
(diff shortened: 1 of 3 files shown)


lib/MetaCPAN/Web/Model/ReleaseInfo.pm
```diff
@@ -165,6 +167,51 @@ sub normalize {
     };
 }
 
+sub groom_repository {
+    my ( $self, $release ) = @_;
+    my $repo = $release->{resources}{repository}
+        or return {};
+
+    $repo = {%$repo};
```
oalders, line 175: What does this do?

lib/MetaCPAN/Web/Model/ReleaseInfo.pm
```diff
@@ -165,6 +167,51 @@ sub normalize {
...
+
+    for my $type (qw(url web)) {
+        my $url = $repo->{$type}
+            or next;
+
+        if (
+            $url =~ m{
+            \A
+            (?:ssh|https?|git)://
```
oalders, line 184: Had you thought about using `URI` here? It supports `git`. https://metacpan.org/pod/URI::git

haarg: I could, but I'm not sure it would really make things any more clear. The regex would end up split into two, and various extra checks would have to be added.

oalders (approved):

PR #2710, opened 2022-07-13
Update layout for metacpan-web

### LINKED ISSUE #2708: Update the layout for https://metacpan.org/
Attached are some potential updates to the layout of https://metacpan.org/. 

The branch [dhogan8/update-layout](https://github.com/metacpan/metacpan-web/tree/dhogan8/update-layout) is still a WIP pending feedback, but here is a list of some potential changes so far:

- Update header and remove icons
- Create a simple footer and remove sponsors
- Move the search to the nav menu except for the homepage
- Update the sidebar and change the screen size it toggles
**Note:** I haven't tested this layout on all pages yet.

about
author
faq medium
home - logged out
home mobile
metadata  medium
mobile slidepanel
news medium
pod
pod renderer
pod2
pod2 contents
tools - logged in

dhogan8:
Closes #2708


oalders:
I'm not clear on the change to `root/static/icons/metacpan-icon-large.png`. In the diff it looks to be fuzzier than the original.

dhogan8:
The main difference is that it has a transparent background. The old one had a white background so it could only be used in certain places on the page that also had a white background. But I do see your point about it being fuzzier. I didn't originally notice this because it isn't too large on the page (with the draft as is, it only appears in the navbar small screen sizes).

oalders:
Ah, ok. If you need the source files, these might be easier to work with: https://github.com/metacpan/metacpan-assets

oalders:
As of today, StickerYou is no longer a sponsor, so we don't need to allow for them in the footer anymore.

haarg:
I haven't had proper time to check everything in terms of the actual look of the design, although I've liked everything in the screenshots. From a technical perspective, I'm happy with the patch.

mohawk2:
Is there any reason not to merge, then deal with any problems that arise?

oalders:
We had talked about deploying this to a staging machine. Having said that we could just deploy this, but it should probably be at a time when someone is also around to deal with any issues that might arise.

oalders:
There's something going on with the staging deployment here. It looks like the JS assets are not being deployed. Should we just be building the static files as part of the Docker build? It's not clear to me how best to handle this, since the static asset compilation is tied up in Plack middleware.

oalders:
For reference, the staging site is at https://web.stage.hc.metacpan.org/ but we still have an issue with static assets not being available, which is messing with the layout.

thibaultduponchelle:
> but we still have an issue with static assets not being available, which is messing with the layout.

Browsing some pages in staging, I don't see any error related to this in the developer console. What are these static assets that are not available? (I thought it was css)

dhogan8:
I don't think the issue will show up as an error. But none of the JS assets are loading on the staging site. Below are two screenshots (one of local development and one of the staging site). None of the JS assets are being loaded properly on the staging site.
staging_site
local_development
:

thibaultduponchelle:
Ok thank you so much for all the details :smiley:

oalders:
@haarg this is ready for another review.

haarg:
I have a few minor issues remaining with this.

  - The way this adds a class to the home page body tag is rather ugly and error-prone. I've filed a PR to fix this: #2744
  - The `root/static/css/jquery.autocomplete.css` file has some whitespace changes - I'd rather leave those out of this PR. This can probably just be dealt with when merging though.
  - The sidebar links have no styling to distinguish them. There are a number of links as well as a number of purely informational entries, and which is which is not always obvious. Needing to hover over the entries (impossible on mobile) seems rather awkward. I wouldn't mind switching the links to be always underlined.

Basically, I'm happy to see this merged. The first two issues I listed can be resolved while merging. The last one would be nice to improve, but doesn't need to hold this up.

oalders:
Merged via #2746. 🎉 🎉 🎉 

Thanks, @dhogan8!

PR #2717, opened 2022-07-28
Add TopBargains to footer sponsors

oalders:
Screen Shot 2022-07-27 at 9 37 52 PM

The formatting is not amazing, but we can fix that as part of the redesign that @dhogan8 has taken on.


ssoriche (approved):
👍

PR #2723, opened 2022-08-19
Drop links to Kritika.IO

thibault-deriv:
Hello :smile: 

Keep the sponsor paragraph for Kritika.io but remove links since service is down for about 1 year and at some point it will link to something unrelated.

### DIFF
--- a/root/about/sponsors.tx
+++ b/root/about/sponsors.tx
@@ -319,14 +319,14 @@ community.
                 </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


oalders:
Thanks! This file is also in that large redesign branch, so we should hold off on this.

thibault-deriv:
Ok :smile:

oalders:
Closed and re-opened to kick off CI builds.

PR #2729, opened 2022-08-22
Apply perlimports

oalders:
- Add perlimports.toml
- perlimports

### DIFF
--- a/lib/MetaCPAN/Web/Test/HTML5/TreeBuilder.pm
+++ b/lib/MetaCPAN/Web/Test/HTML5/TreeBuilder.pm
@@ -4,7 +4,8 @@ use warnings;
 
 use parent 'HTML::TreeBuilder';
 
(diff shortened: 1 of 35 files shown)


lib/MetaCPAN/Web/Test/HTML5/TreeBuilder.pm
```diff
@@ -4,7 +4,6 @@ use warnings;
 
 use parent 'HTML::TreeBuilder';
 
-use MetaCPAN::Web::Test::HTML5::Element::SVG;
```
haarg, line 7: This seems wrong. This module is needed.

oalders: It wasn't clear to me how it's needed in this file. I was only seeing the class name used in an `isa` check.

haarg (approved):

PR #2742, opened 2022-09-23
Make geocode logo fit in better with the existing logos

oalders:
The previous commit were missing some edits to the logo size and
position


ssoriche (approved):
Looks better.

PR #2751, opened 2022-09-30
cope with authors with nothing visible in their name

haarg:
In an author has an "empty" name set, various parts of the UI will not have anything to click on, breaking navigation.

Detect this by checking for word characters in the author name, and falling back to the PAUSE id if there are none. If the API were to provide a display_name, either through a similar check like we are using or as a field that can be filled in, it would be preferred over either the name or the PAUSE id.

### DIFF
--- a/root/base/release.tx
+++ b/root/base/release.tx
@@ -9,7 +9,7 @@
 %%  override breadcrumbs -> {
 <div class="breadcrumbs">
   <span>
-    <a data-keyboard-shortcut="g a" rel="author" href="/author/[% $release.author %]" title="[% $author.asciiname %]" class="author-name">[% $author.name %]</a>
+    <a data-keyboard-shortcut="g a" rel="author" href="/author/[% $release.author %]" class="author-name">[% $author.display_name %]</a>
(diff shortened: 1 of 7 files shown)


oalders (dismissed):
Looks great!

root/base/release.tx
```diff
@@ -9,7 +9,7 @@
 %%  override breadcrumbs -> {
 <div class="breadcrumbs">
   <span>
-    <a data-keyboard-shortcut="g a" rel="author" href="/author/[% $release.author %]" title="[% $author.asciiname %]" class="author-name">[% $author.name %]</a>
+    <a data-keyboard-shortcut="g a" rel="author" href="/author/[% $release.author %]" class="author-name">[% $author.display_name %]</a>
```
oalders, line 12: Any reason for removing the title here? Is that useful for accessibility?

haarg: Generally, the two names will be the same. And I'm not sure that a purely ASCII name is of any value for our use case.

oalders (approved):

Issue #2757, opened 2022-09-30
Use gravatar instead of person icon when user is logged in


oalders:
<img width="335" alt="Screen Shot 2022-09-30 at 13 27 55" src="https://user-images.githubusercontent.com/96205/193324695-5e869826-9343-4ed5-975a-0299d43376f8.png">

Right now the icon turns from black to red after login, which is helpful, but it actually wasn't obvious to me that I was still logged in, because to know this I have to remember the logged-out colour of the icon.

It would be nice to use a gravatar (like GitHub does) to give a clear visual indicator that some is currently logged in.

dhogan8:
@oalders How does the following look?

<img width="1511" alt="Screen Shot 2022-10-01 at 9 16 15 AM" src="https://user-images.githubusercontent.com/52248962/193411373-3810fb53-5c5c-46b4-bd70-4093c3292087.png">
<img width="1512" alt="Screen Shot 2022-10-01 at 9 15 40 AM" src="https://user-images.githubusercontent.com/52248962/193411377-220651ad-6bf2-4a6e-8af9-ba2dfea741ed.png">
<img width="815" alt="Screen Shot 2022-10-01 at 9 16 09 AM" src="https://user-images.githubusercontent.com/52248962/193411375-3cd67025-9058-46c6-9228-5c222262d374.png">
<img width="813" alt="Screen Shot 2022-10-01 at 9 15 49 AM" src="https://user-images.githubusercontent.com/52248962/193411376-428a0a63-e538-46d8-af23-c8990df459fd.png">

oalders:
That looks great!

dhogan8:
Great! I added this to this PR: https://github.com/metacpan/metacpan-web/pull/2759. I'm just struggling a bit with this issue: https://github.com/metacpan/metacpan-web/pull/2759#discussion_r985102751. If it might not be an easy fix to do, I can create a separate PR.

I mainly included this issue and https://github.com/metacpan/metacpan-web/issues/2749 in the same PR since they both touch the navbar icons / buttons.

oalders:
Closing as this has been addressed #2761, #2763 and #2766.

Issue #2752, opened 2022-09-30
Header positions move during page load


haarg:
Various elements of the header move from their initial position after the login check is done, and the user account button in the upper right corner is displayed. This is visually distracting.

Possibly the some elements should be adjusted so that when it is shown, the position doesn't change. Alternatively, it might make more sense to always have the button visible, just not usable or with any content until after the login check.

dhogan8:
@haarg I was able to fix this while working on https://github.com/metacpan/metacpan-web/pull/2759. The icon for being logged in wasn't exactly aligned with the other icons which was causing the weird offset.

dhogan8:
@haarg Is this what you mean when you suggest the following?
> Alternatively, it might make more sense to always have the button visible, just not usable or with any content until after the login check.

https://user-images.githubusercontent.com/52248962/193422691-87cc258a-9910-4660-81c8-b41bc1ea46b5.mov

dhogan8:
@oalders @haarg I think this has been resolved in https://github.com/metacpan/metacpan-web/pull/2759

Issue #2749, opened 2022-09-29
Sidebar is empty on "all releases page"


oalders:
Here's an example: https://metacpan.org/author/OALDERS/releases?size=500

<img width="1565" alt="Screen Shot 2022-09-29 at 15 14 55" src="https://user-images.githubusercontent.com/96205/193122190-0f4e6a24-cb52-47db-b096-abdf80c770d6.png">

dhogan8:
@oalders Thanks! Do you know if the sidebar was always empty for this page? I'm trying to understand if the PR deleted something that was previously there or if it is desired to have something on that sidebar for this specific page.

oalders:
That's a good question @dhogan8. I checked out an earlier commit and it looks like this:

<img width="1569" alt="Screen Shot 2022-09-29 at 16 14 43" src="https://user-images.githubusercontent.com/96205/193132996-c1b1f1ca-a303-4130-b742-e89c985c5bf0.png">

Now I realize why I was confused. I think in this case, it's ok not to have a sidebar, especially for authors who have a list of releases with really long names or abstracts.

dhogan8:
Okay yeah that was on my undocumented list of things to update: Figure out a template that works for pages that don't have or need a sidebar. Like the following:
<img width="1272" alt="Screen Shot 2022-09-29 at 4 19 35 PM" src="https://user-images.githubusercontent.com/52248962/193133692-cfc73517-df0e-43d8-b85e-dbb77c67c9c0.png">

dstar8:
@oalders How do the following look to you?
<img width="1511" alt="Screen Shot 2022-09-29 at 7 02 45 PM" src="https://user-images.githubusercontent.com/29670143/193157095-69f09fe2-4f3e-43e7-8df7-2f6ac151ef7b.png">
<img width="1511" alt="Screen Shot 2022-09-29 at 7 02 16 PM" src="https://user-images.githubusercontent.com/29670143/193157099-57a6b58e-22d7-4026-b3fb-0b8b48203294.png">

dhogan8:
@oalders How do the following look to you?
<img width="1511" alt="Screen Shot 2022-09-29 at 7 02 45 PM" src="https://user-images.githubusercontent.com/29670143/193157095-69f09fe2-4f3e-43e7-8df7-2f6ac151ef7b.png">
[...]

PR #2759, opened 2022-10-01
Add no sidebar class

dhogan8:
- Do not display sidebar and slidepanel icon in header if there is a `no-sidebar` class added to the body tag. As described in https://github.com/metacpan/metacpan-web/issues/2749
- Ensure icons stay in viewport on an actual mobile device (for some reason the nav looks OK in Chrome dev tools, but not when actually viewing on a mobile device)
- ~~Use gravatar for a logged in user icon as described in https://github.com/metacpan/metacpan-web/issues/2757~~ This will be done in a separate PR.
- Stop the shifting of the nav links in the navbar as the page is checking for logged in status: https://github.com/metacpan/metacpan-web/issues/2752
- Round the corner of the button icons and search bar in the navbar
- Add dropdown / chevron icon to the logged in / logged out icons

**Mobile device bug**:

Screen Shot 2022-10-01 at 9 48 52 AM

**Some updated screenshots**:

Screen Shot 2022-10-01 at 7 11 45 PM
Screen Shot 2022-10-01 at 7 11 57 PM
Screen Shot 2022-10-01 at 7 12 11 PM
Screen Shot 2022-10-01 at 7 12 21 PM

### DIFF
--- a/root/base.tx
+++ b/root/base.tx
@@ -147,9 +147,9 @@
                     </ul>
                 </li>
                 %% if ! $current['/'] {
-                    <li class="visible-xs visible-sm">
+                    <li class="icon-slidepanel visible-xs visible-sm">
                       <button data-toggle="slidepanel" data-target=".slidepanel">
-                        <i class="fa fa-bars icon-slidepanel"></i>
+                        <i class="fa fa-bars"></i>
                       </button>
                     </li>
                 %% }
--- a/root/base.tx
+++ b/root/base.tx
@@ -125,7 +125,9 @@
                 <form action="/account/logout" method="POST" id="metacpan-logout"></form>
                 <li class="dropdown logged_in" style="display: none;">
                     <button type="button" class="dropdown-toggle" data-toggle="dropdown">
-                      <img class="logged-in-gravatar" src="[% gravatar_image($author) %]">
+                      <!--Add the below back in once we have a way to store a logged in user's gravatar-->
+                      <!--<img class="logged-in-gravatar" src="[% gravatar_image() %]">-->
+                      <i class="fa fa-user button-fa-icon" aria-hidden="true"></i>
                       <i class="fas fa-chevron-down"></i>
                     </button>
                     <ul class="dropdown-menu">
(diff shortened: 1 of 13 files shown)


root/base.tx
```diff
@@ -115,10 +115,17 @@
...
+                      <button data-toggle="slidepanel" data-target=".slidepanel">
+                        <i class="fa fa-bars button-fa-icon"></i>
+                      </button>
+                    </li>
+                %% }
                 <form action="/account/logout" method="POST" id="metacpan-logout"></form>
                 <li class="dropdown logged_in" style="display: none;">
                     <button type="button" class="dropdown-toggle" data-toggle="dropdown">
-                      <i class="fa fa-user" aria-hidden="true"></i>
+                      <img class="logged-in-gravatar" src="[% gravatar_image($author) %]">
```
dhogan8, line 128: I'm a bit stuck on figuring out the right syntax for `[% gravatar_image($author) %]`. I'm not sure what to replace $author with in order to get the gravatar of the logged in user. I temporarily added this in just to be able to visualize what the gravatar _would_ look like.

oalders: Right. So, I think `$author` will refer to the author of a module on a page that's being viewed rather than the profile of a logged-in user (who in some cases may not even be an author).

I think what we might want to do is update the login logic so that the user's gravatar_url is always available in the session after login or have some other way to make it available for every page request. So, this particular fix might need to live in a different branch until that back end code change has taken place.

haarg: We'll need to add the appropriate data to the login_status request. It's probably best to put it in a separate branch/PR.

One challenge here is that for authors, we use the `@cpan.org` email alias for the gravatar. Not all logged in users have an associated PAUSE account. So we'll still need a default user icon.

dhogan8: That makes sense. For now, I added back in the red icon color.

Screen Shot 2022-10-02 at 8 16 20 AM
Screen Shot 2022-10-02 at 8 16 56 AM

haarg (approved):

PR #2761, opened 2022-10-03
Provide avatar though login_status end point

haarg:
This provides the user's avatar in the login_status end point, to allow it to be displayed on the account button.

Hard coding the size in this location is not ideal, but improving that can be left to later.


oalders (approved):
\o/

PR #2763, opened 2022-10-04
update Gravatar handling to provide a template string to login_status

haarg:
Rather than hard coding a size, or trying to accept it as a parameter, return a template string for the avatar, and let the client format it.

The Gravatar::URL module won't allow a string like this, but the module is doing basically nothing for us, so we can just remove it.


oalders (approved):

oalders:
@haarg we have a merge conflict in the `cpanfile.snapshot`.

oalders (approved):

PR #2765, opened 2022-10-04
remove ServerStatus::Lite plugin that we don't use for anything

haarg:


oalders (approved):
Nice. This has been around for many years, but I don't know if I've ever used it. 🧹

PR #2767, opened 2022-10-04
Shorten nav height and specify specific styling for Safari

dhogan8:
- There is some default styling that gets applied in the Chrome and Firefox browsers, but not in Safari (particularly on the an actual mobile device). I explicitly set some of that styling so that the spacing of the navbar icons on Safari look similar to how they look in the other two browsers.
- The navbar height looked a tad too tall on a mobile device, so I shortened it a bit. This should enhance the mobile experience a bit by providing more room for content reading. Shortening the height wasn't too noticeable on a larger desktop, so I kept the same height for both mobile and desktop.

To confirm everything looked ok, here are screenshots with the changes directly from my iPhone in Safari (ignore the white lines around the nav icons, those are just a result of modifying the elements using the inspector tool).

IMG_9865

IMG_9866

### DIFF
--- a/root/static/less/global.less
+++ b/root/static/less/global.less
@@ -548,7 +548,7 @@ form#metacpan_logout {
 }
 
 * {
-  scroll-margin-top: 80px;
+  scroll-margin-top: 60px;
 }
 
 .no-sidebar .nav-list-container,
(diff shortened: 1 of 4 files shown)


dhogan8:
@haarg This should fix that weird icon issue in Safari mobile.
Screen Shot 2022-10-04 at 7 11 18 PM

haarg (approved):

Issue #2768, opened 2022-10-05
Logged in drop-down never appears


openstrike:
**Describe the bug**
When logged in to MetaCPAN the dropdown menu from the avatar in the very top right never appears. Clicking on the avatar toggles the `aria-expanded` value on the button and the class on the enclosing `li` but the actual menu items never show.

**To Reproduce**
1. Login to MetaCPAN
2. Visit any page, eg https://metacpan.org/pod/Pepper
3. Click on your avatar in the top right
4. ... nothing visible happens

**Expected behavior**
The drop-down menu should appear on one click and disappear on the next click.

**Screenshots**
![metacpan-no-menu](https://user-images.githubusercontent.com/4331674/194050822-86f577f3-8692-48e6-90bd-a153861839be.png)

**Desktop (please complete the following information):**
 - OS: Linux 64bit
 - Browser Firefox 105.0.1 and Palemoon 31.3.0.1

haarg:
Fixed by #2769

PR #2769, opened 2022-10-05
Fix logged in status icon dropdown

dhogan8:
In https://github.com/metacpan/metacpan-web/pull/2767, I introduced a bug where I set `overflow-y: hidden` for the `navbar-right` in order to get the nav icons to show in the viewport on Safari mobile. I was so focused on fixing that, that I didn't realize the dropdown menu won't appear when the logged in / out icon is clicked.

Fixed issue:
IMG_9868

### DIFF
--- a/root/static/less/nav.less
+++ b/root/static/less/nav.less
@@ -1,3 +1,13 @@
+// For some reason in Safari, the .clearfix() styling
+// is causing the icons to be outside of the viewport.
+.nav {
+  &:before,
+  &:after {
+    content: "" !important;
+    display: block !important;
+  }
+}
+
 .navbar {
     background-color: @gray-darker;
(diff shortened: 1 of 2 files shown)


root/static/less/nav.less
```diff
@@ -1,3 +1,13 @@
+// For some reason in Safari, the .clearfix() styling
+// is causing the icons to be outside of the viewport.
+.nav {
+  &:before,
+  &:after {
+    content: "" !important;
+    display: block !important;
+  }
+}
```
dhogan8, line 9: I'm still getting used to using the dev tools on a mobile device, but this seemed to do the trick.

haarg (approved):

haarg:
Fixes #2768

PR #2770, opened 2022-10-05
Image cleanups

haarg:
Remove some unused png files, optimize the rest, and replace the ugly grid icon with an SVG file.


oalders (approved):
🧹

PR #2771, opened 2022-10-05
cleanups of less code and other misc

haarg:
This cleans up some bits in the less code. Some whitespace cleanups, removing some unused code, and removing some needless use of mixins. Also removes a problematic use of the extend feature that would be hard to adapt to any other CSS templating system.

### DIFF
--- a/root/base/release.tx
+++ b/root/base/release.tx
@@ -12,7 +12,7 @@
     <a data-keyboard-shortcut="g a" rel="author" href="/author/[% $release.author %]" class="author-name">[% $author.display_name %]</a>
   </span>
   <span>&nbsp;/&nbsp;</span>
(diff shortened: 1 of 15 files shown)


dhogan8 (approved):
Looks good!

One question for my own learning - Is there a preference between using 2 spaces vs. 4 spaces for indenting? I saw a little bit of both throughout the codebase and wasn't sure if there was a preference between the two.

haarg:
The perl code is all using 4 spaces, and most of the less and javascript code is using 4 spaces. And the `.editorconfig` file is configured to use 4 spaces. So that's what I think we should standardize on.

The `.tx` and `.html` files are configured to use 2 spaces, and that's generally what they are doing. I'm fine with sticking with that. We could change that if we wanted, although reindenting the templates isn't as straightforward as the code.

PR #2772, opened 2022-10-05
split table of contents in backend code

haarg:


haarg:
This is a work in progress branch for splitting the table of contents out server side rather than in javascript. It's currently functional, but I wrote this a bit ago and need to review it again myself.

PR #2773, opened 2022-10-05
Use flex instead of inline

dhogan8:
Original bug:
Screen Shot 2022-10-05 at 5 30 23 PM

Fixed version:

Screen Shot 2022-10-05 at 5 21 00 PM


dhogan8:
Screenshot confirming the dropdown still works:
Screen Shot 2022-10-05 at 5 46 58 PM

oalders (approved):

PR #2774, opened 2022-10-06
Use Deriv home page instead careers

thibault-deriv:
Preferred link to share for Deriv :smiley:


oalders:
Just to be clear, when I initially spoke with Deriv about the logo, they provided the careers link to me. Have they changed their mind?

thibault-deriv:
@tom-binary Do you agree with this change?

thibault-deriv:
:man_shrugging: I will close this, it is not important

PR #2779, opened 2022-10-07
use github avatar when available for users without a pause account linked

haarg:


oalders (approved):
Sorry. Somehow I missed this.

PR #2782, opened 2022-10-10
fix month in changelog parser calculations

### LINKED ISSUE #2780: Distro "changes for version XX" summary gives month off-by-one
**Describe the bug**
The month in the distro-level summary is off by one for at least PDL-IO-Touchstone-1.006. I surmise that while MetaCPAN::Web::Model::API::Changes::Parser is parsing month names into a zero-based integer, the ISO 8601 formatter in `root/release.tx`, `datetime($rel.date).to_ymd()` seems to be expecting a one-based integer for month. That comes from MetaCPAN::Web::Role::Date, which uses `strftime("%F")` (which from MetaCPAN::Web::View::Xslate::Bridge appears to be getting added as a role to DateTime). While `struct tm` from `time.h` uses zero-based for the month, it seems there is an expectation mismatch here.

There's a few explicit uses of `datetime($release.date).strftime("%F");` (e.g. `root/inc/release_table.tx`) which should probably use `to_ymd` as well.

**To Reproduce**
- https://metacpan.org/release/EWHEELER/PDL-IO-Touchstone-1.006 shows the summary being for "Changes for version 1.006 - 2022-09-07" (month 09 = September)
- https://metacpan.org/release/EWHEELER/PDL-IO-Touchstone-1.006/source/Changes shows the date spelled out "Fri, 07 Oct 2022 21:41:30 -0700" (October)

**Expected behavior**
The months should not be off by one.

**Screenshots**

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**
None

haarg:
When parsing the date from a change log, the short month was being matched, and the index of the match was used as the month. Since the index starts at 0, this resulted in the month being off by one.

Add one to the index before using it in our short month to number hash.

Fixes #2780

### DIFF
--- a/lib/MetaCPAN/Web/Model/API/Changes/Parser.pm
+++ b/lib/MetaCPAN/Web/Model/API/Changes/Parser.pm
@@ -4,7 +4,7 @@ use Moose;
 use version qw();
 
 my @months = qw( Jan Feb Mar Apr May Jun Jul Aug Sep Oct Nov Dec );
-my %months = map +( $months[$_] => $_ ), 0 .. $#months;
+my %months = map +( $months[$_] => $_ + 1 ), 0 .. $#months;
 my $months = join '|', @months;
 
 our $W3CDTF_REGEX = qr{


oalders (approved):
📆

PR #2786, opened 2022-10-14
Keep CPAN/BackPAN links together in author sidebar

oalders:


haarg:
This change on its own seems fine, but I think we should revisit the cpan vs backpan links in general.

haarg (approved):

oalders:
Yes, it would be a good idea to think about how to present the links. I just found it weird that we had inserted the permissions link *between* the CPAN links. Also, I was probably the person who did that. 🤷🏻‍♂️

PR #2790, opened 2022-10-18
replace all logos with one svg file

haarg:
Add a logo svg file that can be used for nearly all of our uses. By default, it displays the classic logo we've been using everywhere until recently, with grey text and red dots. But it includes the dots as a distinct symbol that can be selected, as well as exposing a CSS variable that can control the text color.

This means we can remove basically all of the other logos, and only use one file.

The dots.svg file could also be converted to reference the main logo file, but it didn't seem worth it.


haarg:
Might be worth testing this change on a few other browsers, including mobile.

oalders (dismissed):

oalders (dismissed):

oalders (approved):

PR #2796, opened 2022-10-26
improve display when overflow scrolling

haarg:
When scrolling past the top of the page, the header would move down from the top of the display, and show white behind it. And scrolling past the bottom would also show white.

Change the positioning of the header from sticky to fixed. This means it will always be attached to the top, even with overflow scrolling. sticky is more useful for things further down the page that you want to attach to the top once you've scrolled past them. Changing to fixed also results in the header not taking up any space in the layout. Adjust the top margin of the page content to accomodate for this. This is makes rules match for the desktop and mobile layout.

We can set the background color browsers use for overflow scrolling with the background on the html element. We're already setting the body background to white, so the html background won't be visible anywhere in the normal content. Browsers only seem to use the background color for overflow scrolling, so we can't use gradients or anything else. But since we've fixed the header to the top of the page, we only need to worry about matching the color of the footer.

### DIFF
--- a/root/static/less/footer.less
+++ b/root/static/less/footer.less
@@ -5,6 +5,10 @@
   padding: 30px;
   width: 100%;
 }
+// for overflow scrolling
+html {
+    background-color: #e9e9e9;
+}
 
 .footer-container {
   display: grid;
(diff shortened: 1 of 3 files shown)


dhogan8 (approved):
✨

PR #2800, opened 2022-10-31
Refactor handling of file data

haarg:
This adds a role for cleaning up file data coming from the API. Fixes the MIME type, and sets some flags relating to the available modules.

Uses the new more consistent MIME type directly in some places, rather than having matches on the file extension in various places around the codebase.

Also renames a bunch of variables to file rather than module, for better consistency.


mickeyn (approved):
LGTM

PR #2823, opened 2022-11-08
Fix handling of non-UTF8 source code

haarg:
When Text::Xslate is given data with the utf8 flag off, but with valid
non-strict utf8 sequences in it, those sequences will be decoded. That
will then produce errors when attempting to encode it for output. We
only really run into this problem when showing the source of a file,
since those may have arbitrary bytes.

Force the source content given to Xslate to always have the utf8 flag on,
so that Text::Xslate inserts the non-decoded "characters" directly into
the output. They will essentially be mojibake, but this is acceptable
since the original file had badly encoded data in it.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Source.pm
+++ b/lib/MetaCPAN/Web/Controller/Source.pm
@@ -68,6 +68,12 @@ sub view : Private {
         } );
     }
     elsif ( exists $source->{raw} ) {
+
+        # hack to work around Xslate. utf8-flagged content will be encoded to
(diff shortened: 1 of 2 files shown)


oalders (approved):

oalders:
```
Can't call method "info" on an undefined value at /metacpan-web/lib/MetaCPAN/Web/Model/API.pm line 196.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 2.
t/encoding.t ............................. 
```

oalders (approved):

PR #2841, opened 2022-11-28
changes parser: fix comparing versions that are in different forms

haarg:
When finding the changes for a version v1.2.3, where the change log contains "1.2.3", we need to be comparing the versions as version objects rather than strings. But we still need to prevent version.pm from trying to inflate the other version it is being compared to, because that already failed. Force a stringificiation on both versions if either failed to parse.


mickeyn (approved):

oalders:
Needs tidying.

oalders (approved):

PR #2857, opened 2023-01-09
rewrite changes handling for rss feed

haarg:
The previous code would batch requests in the controller, manually generate URLs in the model, and had odd inputs and outputs.

Rewrite the changes model by_release method to accept a simple list of release paths, handling batching internally, and return a simple hashref. Doing the batching internally makes it easy to add concurrency to the requests.

Also fix the version regex to understand v prefixes.

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Feed.pm
+++ b/lib/MetaCPAN/Web/Controller/Feed.pm
@@ -29,26 +29,16 @@ sub recent : Private {
     $c->forward('/recent/index');
     die join "\n", @{ $c->error } if @{ $c->error };
 
-    my %changes_index;
-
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Web/Controller/Feed.pm
```diff
@@ -31,24 +31,16 @@ sub recent : Private {
 
     my %changes_index;
```
oalders, line 32: It looks like after these changes, this variable is no longer being used.

oalders (changes requested):
Looks great. Just need to remove that one variable to get the tests passing.

oalders (approved):

PR #2894, opened 2023-04-28
Switch Markdown rendering from Text::MultiMarkdown to CommonMark

### LINKED ISSUE #2312: Markdown not rendering correctly
Please compare 

https://github.com/marioroy/mce-perl/blob/master/README.md

vs

https://metacpan.org/source/MARIOROY/MCE-1.872/README.md

The 3x ` seem to have problems.

ilmari:
Custom rendering function to add `id` attribute to headings.

Fixes #2312


cpanfile.snapshot
```diff
@@ -221,32 +221,6 @@ DISTRIBUTIONS
...
-      Encode 0
-      Exporter 0
-      ExtUtils::MakeMaker 6.17
-      File::Spec 0.80
-      JSON::PP 2.27300
-      Scalar::Util 0
-      perl 5.008001
-      strict 0
-      version 0.88
-      warnings 0
```
haarg, line 249: Can you reduce the changes to `cpanfile.snapshot` to only include addition of `CommonMark` and `Devel-CheckLib`? And add `CommonMark`.

haarg (changes requested):
Generally looks good.

We'll need to wait to merge this until we can change our deployments to include libcmark. It also looks like it needs to get installed in CI.

cpanfile.snapshot
ilmari: I've removed the spurious changes, but where do you think I should add `CommonMark` besides the places I already have?

oalders:
The docker container has no `sudo`, so you can omit that from the command.

haarg (approved):

haarg:
The content of this looks good. I've marked it as draft so that it doesn't get merged yet, until we can get the front end switched over to our new systems.

haarg (approved):

PR #2939, opened 2023-09-23
all_by_author: called with the wrong params order (GH#1126)

mickeyn:
match call params order with sub:

> sub all_by_author {
>     my ( $self, $pauseid, $page, $page_size ) = @_;


haarg (approved):

oalders (approved):

nrdvana:
This broke the author releases pagination... #2941

PR #2955, opened 2023-11-24
Add colors for release name in recent page

### LINKED ISSUE #2954: Improve recent page with indication of type of release (stable/testing/unstable/cpan)
**Describe the feature**
Improve recent page https://metacpan.org/recent with information about release type.
I think that could be some icon before the river icon or by the color of the link (same as in the distribution page)

**Which problem does it solve?**
When I look at the recent page there is a mix of secret releases (status => cpan), public releases (status => latest), development releases (meta.release_status => unstable), and testing releases (meta.release_status => testing).
This is not fine for the user.

**Expected behavior**
When I look at the recent page, I can identify releases, that I need.

**Volunteering**
I could implement it.

michal-josef-spacek:
We have different types of releases (development, testing, only for cpan) and there were no visible identification of these.

Fixes #2954

### DIFF
--- a/root/inc/release_table.tx
+++ b/root/inc/release_table.tx
@@ -27,7 +27,7 @@
     %%  }
         <tr>
           <td class="river-gauge" sort="[% $release.river.total || 0 %]">[% include inc::river_gauge { distribution => $release.distribution, river => $release.river } %]</td>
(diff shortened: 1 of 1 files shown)


haarg:
There is a slight problem with this, which is that the table being modified is used in a few other places, like the author and prereqs pages. These don't all provide the needed meta fields to create the proper classes. Those should probably be added before merging this so we're certain it looks acceptable.

haarg:
metacpan/metacpan-api#1150

haarg (approved):

PR #2977, opened 2024-01-22
add regeneration to static-files.t test

haarg:
find may include things like swap files, and sort is locale dependent. It's more reliable to implement the regeneration ourselves.


oalders:
@haarg needs a rebase?

oalders (approved):

PR #3036, opened 2024-04-28
build assets with esbuild, with docker compose for running with auto-reload

haarg:


Dockerfile
```diff
@@ -1,8 +1,95 @@
-FROM metacpan/metacpan-base:latest
+################### Asset Builder
 
-ARG CPM_ARGS=--with-test
+FROM node:latest AS build-assets
```
oalders, line 3: I wonder if we want to pin to a version so that we don't get any surprises?

oalders (approved):

PR #3048, opened 2024-05-03
oalders/precious

oalders:
- **Add precious config**
- **Add script to install precious in Docker**


.github/workflows/test.yml
```diff
@@ -83,3 +88,17 @@ jobs:
         run: cover -report codecov
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+      - name: Install precious
```
haarg, line 91: Don't we want this to be in a separate job, so it's shown as a separate status check?

haarg (approved):

Issue #3062, opened 2024-05-12
"add to favorites" produces 403 error


niceperl:
**To Reproduce**
1. Login
2. Search a good distribution you like
3. Click on "add to favorites" (number with favourites count)
4. Get the error: "Forbidden, You don't have permission to access this resource."

**Expected behavior**
Increase the favourites count

Issue #3065, opened 2024-05-12
Can't ++ any distribution with "Forbidden" error


rwp0:
When I want to `++` I get this [error](https://metacpan.org/account/favorite/add?remove=0&release=Authen-TOTP-0.1.0&author=TCHATZI&distribution=Authen-TOTP):

![image](https://github.com/metacpan/metacpan-web/assets/68062695/1384c927-eb2c-48e6-8c4b-ebe8367313cd)

It's something recent, previously I could `++` normally.

I think it's has something to do with [Identities](https://metacpan.org/account/identities):

![image](https://github.com/metacpan/metacpan-web/assets/68062695/8b59ee9d-a6e1-4be6-8305-0c04b748ff2b)

I tried to disable and enable few combinations to remedy this issue with no luck.

haarg:
Dupe of #3062

PR #3066, opened 2024-05-13
use vendor libraries from npm and various fixes

haarg:
This removes the vendored libraries we had, and instead uses them from npm.

It also fixes various problems that were introduced with the conversion to esbuild. With that conversion, the jQuery object and other functions were no longer globally available. This broke various bits of inline code.

This removes all of the inline code, like `onclick` handlers.


haarg:
This should fix #3062 and #3065

root/static/js/cpan.js
```diff
@@ -325,126 +227,97 @@
...
+            alert('Please sign in to add favorites');
+        });
+    }
 
-    $(document).on("dragenter", function(event) {
-        renderer.addClass("dragging");
-    });
+    for (const sel of document.querySelectorAll('.select-navigator')) {
+        sel.addEventListener('change', () => {
+            document.location.href = sel.value;
```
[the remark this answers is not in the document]
haarg: In the way this is implemented, it's no more vulnerable than an `<a href="...">`.

oalders:
@dhogan8 if you have a moment, can you take a *quick* look at the front end changes? The diff is too large for a human to handle, but even if you spent 5 or 10 minutes on it, that would be helpful.

Dockerfile
```diff
@@ -41,7 +41,7 @@ WORKDIR /app/
 COPY cpanfile cpanfile.snapshot ./
 RUN \
     --mount=type=cache,target=/root/.perl-cpm,sharing=private \
-<<EOT /bin/bash -euo pipefail
+<<EOT
```
oalders, line 44: Wondering why the bash strict mode is getting removed?

oalders (approved):

Dockerfile
haarg: It's already set with a `SHELL` command. I've gone back and forth on how I wanted to handle this, but for now the `SHELL` command seems fine.

PR #3071, opened 2024-05-13
Try to normalize the github workflows

haarg:
Try to get closer to a consistent setup for GitHub workflows. Use actions that can be shared across all of our repos and try to use more 'official' actions.


haarg:
Along with all of the other changes, this also separates out the tidy checks into a separate action run. #3048 also tries to address this, so whichever PR is merged second will need some adjustment. I included it in this PR because the `codecov` coverage report was causing failures and is deprecated, while the replacement `codecovbash` was failing due to a strange interaction with `Test::Vars` which was being used in the tidy test.

oalders:
This one failed the formatting check, so I think it still needs a tweak. We can remove all references to tidyall at this point, I think.

.github/workflows/test.yml
```diff
@@ -10,25 +10,40 @@ on:
...
+    steps:
+      - name: Build test image
+        id: docker-build-test
+        uses: docker/build-push-action@v5
+        with:
+          target: test
+          push: false
+          load: true
+      - name: Run formating check
+        run: docker run -i ${{ steps.docker-build-test.outputs.imageid }} tidyall --all --check-only --no-cache
```
oalders, line 25: We'll need to tweak this to use precious

oalders (approved):

PR #3072, opened 2024-05-13
fix various onclick handlers by explicitly attaching functions to window

### LINKED ISSUE #3062: "add to favorites" produces 403 error

haarg:
This fixes #3062 as well as several other unreported issues.

### DIFF
--- a/root/static/js/cpan.js
+++ b/root/static/js/cpan.js
@@ -571,3 +571,4 @@ function favDistribution(form) {
     });
     return false;
 }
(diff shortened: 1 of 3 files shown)


haarg:
This is a lower impact alternative to #3066, which is large and harder to review.

PR #3078, opened 2024-05-16
provide opengraph metadata in addition to twitter card data

haarg:


oalders (dismissed):
Thank you! This will be a very nice improvement.

oalders (approved):

PR #3084, opened 2024-05-20
split codecov uploading into separate repo

haarg:
Some PRs run with lowered permissions that don't have access to secrets. This is generally good, but means they don't have a token to upload to Codecov. Split out the uploading to a separate workflow that is triggered by the completion of the test workflow.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,5 @@
 ---
-name: test
+name: Test
 on:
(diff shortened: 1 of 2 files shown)


oalders:
We'll need to resolve the conflicts here.

haarg:
I'll close this for now. I didn't get this working at all, so it needs a full redo.

Issue #3102, opened 2024-06-11
Sponsor page has gone AWOL


oalders:
**Sponsor logos and images are being filtered out on render**

**To Reproduce** Steps to reproduce the behavior:

1. Go to <https://metacpan.org/about/sponsors>

<img width="928" alt="Screenshot 2024-06-10 at 8 45 49 PM" src="https://github.com/metacpan/metacpan-web/assets/96205/06bf9bc6-9388-4dcc-9faf-834da227a8d0">

The issue looks to be in the template rendering:

```html
<h2 id="our-current-and-previous-big-sponsors">Our current and previous big sponsors</h2>
--
  | <!-- raw HTML omitted -->
  | </main>
  | <div class="content-pagination">
  | </div>
  | </div>
```

PR #3103, opened 2024-06-11
allow raw HTML to be used in markdown

haarg:
We do our own filtering of any user controlled markdown files, so it's safe to allow arbitrary HTML to be used. And we want raw HTML to be usable for user markdown and especially for our own markdown.

Enable the OPT_UNSAFE flag to allow raw HTML.


haarg:
Fixes #3102

oalders (approved):

PR #3137, opened 2024-07-21
Document how to install CommonMark on newer Macs

oalders:


README.md
```diff
@@ -48,6 +53,31 @@ OPENSSL_PREFIX="/usr/local/Cellar/openssl@1.1/1.1.1q" carton install
 You may need to check `/usr/local/Cellar/openssl@1.1` to find the latest
 installed path on your system.
 
+#### cmark
+
+If you like, you can install the `cmark` C libraries via Homebrew:
+
+```bash
+brew install cmark
```
haarg, line 61: I think this may be better left as part of the `brew install carton` command. The elaboration added in this PR doesn't change anything about how cmark is installed, just how it's used. And it should generally be installed before running carton.

Linking against the version specific paths shouldn't be needed and will be fragile. Upgrades could lead to breaking your build. So I'd rather list it as a fallback in case the normal path didn't work rather than the expected path.

haarg:
So I've figured out what allowed this to work fine for me on my system. I have `LIBRARY_PATH` and `CPATH` set to allow compiling against the homebrew libraries.

`LIBRARY_PATH=/opt/homebrew/lib CPATH=/opt/homebrew/include cpm install -g CommonMark` is likely a better command to recommend.

oalders:
@haarg this is ready for another review.

haarg (approved):

PR #3156, opened 2024-08-24
Return 404 for invalid path: /release//File-Stat-Convert

### LINKED ISSUE #3155: Return a 404 for /release//Text-ETSV
`https://metacpan.org/release//Text-ETSV` (and others) return a 500 (it's being hit by various bots)

Error is `MetaCPAN::Web::Controller::Release->release_view "dist or author and release must be provided.."`

Should return a 404 or maybe a 400 as the URL is actually invalid?

ranguard:
Return 404 for invalid paths, e.g. /release//File-Stat-Convert instead of a 500 - fix #3155

Please could you run perl tidy on this - I don't have the Perl setup


ranguard:
#3157 is better

PR #3157, opened 2024-08-25
Return 404 for releases missing author, e.g. /release//File-Stat-Convert instead of a 500

### LINKED ISSUE #3155: Return a 404 for /release//Text-ETSV

haarg:
Fixes #3155

### DIFF
--- a/lib/MetaCPAN/Web/Controller/Release.pm
+++ b/lib/MetaCPAN/Web/Controller/Release.pm
@@ -1,13 +1,15 @@
 package MetaCPAN::Web::Controller::Release;
 
 use Moose;
-use namespace::autoclean;
+use namespace::autoclean -except => [qw(NonEmptyStr)];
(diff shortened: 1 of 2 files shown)


haarg:
This is an alternative to #3156. Instead of checking and rejecting an empty author, this adds type constraints to the routing. A path that we don't want to handle will just fall out as unroutable and thus give a 404. This style should nicer to use in more places, rather than having to put explicit checks and 404 redirects all over the code.

ranguard (approved):
Nice, it is what I wanted to do but didn’t know how!

PR #3201, opened 2024-10-09
Fix alignment of magnifying glass in search input

tessarin:
The magnifying glass icon is now centered vertically without relying on its absolute size. This ensures proper centering, regardless of the icon's dimensions.

Original:
Screenshot 2024-10-09 at 12 18 51

Fixed:
Screenshot 2024-10-09 at 12 18 38

### DIFF
--- a/root/static/less/search.less
+++ b/root/static/less/search.less
@@ -30,9 +30,10 @@
 .search-group .fa-search {
     color: @input-color;
     font-size: 20px;
-    left: 10px;
+    left: 18px;
     position: absolute;
-    top: 23%;
+    top: 50%;
+    transform: translateY(-50%);
 }
 
 .navbar-nav .search-form {


oalders:
Thanks! Can you confirm that this works in various other dimensions? If you could include a screen shot from something like Chrome's developer tools that shows what this looks like on mobile, that would be helpful.

tessarin:
@oalders I does — I did in fact check that. Bellow is a screenshot of Safari dev tools (on a rather extreme narrow width), but the result is the same in Chrome.

Screenshot 2024-10-09 at 13 27 08

tessarin:
Now that I'm looking at it again, I think moving one extra pixel to the right would make it even better in terms of visual weight centering.

17px
Screenshot 2024-10-09 at 13 33 22

18px
Screenshot 2024-10-09 at 13 34 07

tessarin:
Screenshot 2024-10-09 at 13 42 28

oalders (approved):
Thanks, @tessarin!

PR #3224, opened 2024-11-04
pod2html: fix rendering in chrome

### LINKED ISSUE #3223: Pod Renderer does not work well
**Describe the bug**

Pod Renderer does not work well.

https://metacpan.org/pod2html

**Screenshots**

image

haarg:
Elements that are part of a DocumentFragment don't seem to have a working outerHTML property in Chrome.

Fixes #3223.


yuki-kimoto:
Thanks.

PR #3231, opened 2024-11-18
Fix syntax highlighting alignment and =cut handling

haarg:


mauke:
Related: https://github.com/Perl/perl5/issues/22759

PR #3239, opened 2024-11-25
Remove links to author backpan dir as we want to stop supporting that

ranguard:


oalders:
@ranguard other than needing a rebase, is this ready to deploy?

ranguard:
@oalders I believe so!

oalders (approved):

PR #3293, opened 2025-03-09
cleanup mailmap

mickeyn:


haarg:
I'm not sure I understand what this change is meant to do. The mailmap entry being removed here rewrites the apparent author to use the email `mickey@cpan.org`, as there are several commits from 2016 with the wrong email `micha.nasriachi@booking.com`.

mickeyn:
@haarg ah, OK. I just wanted to remove references to this email from being available on Google search. 
I'll close it.

Issue #2907, opened 2023-05-12
Would be nice to show the number of search hits at the top of the page


ingydotnet:
https://metacpan.org/search?size=500&q=api shows

> 6,568 results (0.805 seconds)

at the bottom of the page (of 500 items).

Issue #1686, opened 2016-05-03
Pagination UI is visually confusing


haarg:
Having the results per page use basically the same UI as the pages and directly below it is visually confusing.  I always have to pause briefly to read the text, when usually a pagination UI should be self evident.

I think a drop down list to the right would be a better choice for the results per page.  It is unlikely to be changed often, so having the options available within one click isn't needed.

oalders:
I think we can close this as we no longer have a UI element for choosing results per page.

PR #3325, opened 2025-05-01
show search hits at page top

JordanEdward02:
Adds the number of search hits and some formatting to the top of the search page for #2907

### DIFF
--- a/root/search.tx
+++ b/root/search.tx
@@ -4,6 +4,16 @@
 %%  after content_classes -> { ' search-results' }
 %%  override content -> {
   <h3 class="search-results-header">Search results for "[% $search_query %]"</h3>
+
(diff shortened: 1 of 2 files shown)


guest20:
Can you also change the "results per page" to not look exactly like the pagination controls when I'm not wearing my glasses?

JordanEdward02:
> Can you also change the "results per page" to not look exactly like the pagination controls when I'm not wearing my glasses?

I'm not sure I understand what you mean. Could you raise a separate issue with some images to highlight the problem please?

guest20:
There are two already - https://github.com/metacpan/metacpan-web/issues/1686 https://github.com/metacpan/metacpan-web/issues/2218

It's about this control, which is a pagination widget being used for non-pagination related purposes: 
image

Issue #3278, opened 2025-02-12
BackPAN link not pointing to BackPAN


eserte:
When on an author page, e.g. https://metacpan.org/author/TONYC , then the "BackPAN directory" link shows to the same page, not to the author's backpan page.

ranguard:
We are no longer supporting directory listings (because we are moving the backpan to an object store).

At some future date we may work building one but it is not priority for now.

The `BackPAN Directory` link was removed in https://github.com/metacpan/metacpan-web/pull/3239

PR #3327, opened 2025-05-01
Proposed patch for issue #3278

manwar:
Hi Team.

Here is another issue dealt with.

    BackPAN link not pointing to BackPAN
    https://metacpan.org/author/TONYC

Please review when you get time.

Many Thanks.

Best Regards.
Mohammad


haarg:
I believe this is better handled in the API. We also want to link to metacpan's mirror as the backpan.

Here is an alternate fix for this: metacpan/metacpan-api#1375

oalders:
@manwar can you confirm that the merged PR on the API side fixes this?

oalders:
Closing as our behaviour around BackPAN links has changed.

PR #3332, opened 2025-05-02
Proposed patch for issue 3108

manwar:
Hi Team,

Please review the small patch when you get time.
  
    Added attribute "alt" with value "Activity Graph" to the img element.
    https://metacpan.org/pod/Test2

Many Thanks.
Best Regards,
Mohammad

### DIFF
--- a/root/inc/activity.tx
+++ b/root/inc/activity.tx
@@ -1,4 +1,4 @@
 <div class="activity-graph">
-    <img src="[% $url %]?res=month" />
+    <img src="[% $url %]?res=month" alt="Activity Graph"/>
     <div class="comment">24 month</div>
 </div>


root/inc/activity.tx
```diff
@@ -1,4 +1,4 @@
 <div class="activity-graph">
-    <img src="[% $url %]?res=month" />
+    <img src="[% $url %]?res=month" alt="Activity Graph"/>
```
guest20, line 2: This is a bad alt text. It just repeats the heading directly above it without replacing the information presented in the graph.

One option would be to count the number of groups in the graph and describe how high each of them are. 

The other choice is to alt it as "Changes" and link to the whole thing to the `Changes` file

oalders (approved):

PR #3336, opened 2025-05-03
Upgrade bootstrap to 5.3.5 and remove ttip package

JordanEdward02:
Removing the ttip package is just a start at
removing jquery as a dependency. The browser
can handle titles itself to create popups so I
have just migrated them to use that instead.

We still have a couple of jquery uses in the code:
 - Keyboard shortcuts modal popup
 - Dropdown menu for the login and profile button

These are not trivial to remove unlike ttip so
I will not include them in this branch

Examples of the revised tooltips:

Screenshot from 2025-05-03 13-37-09
Screenshot from 2025-05-03 13-36-57

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "dependencies": {
         "@fortawesome/fontawesome-free": "^6.7.2",
-        "bootstrap": "^3.4.1",
+        "bootstrap": "^5.3.5",
         "brush-cpp": "^4.0.1",
         "brush-diff": "^4.0.0",
         "brush-javascript": "^4.0.0",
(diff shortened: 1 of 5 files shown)


haarg:
This updates the Bootstrap version, but doesn't do anything to adapt to the new version.

To start with, we currently include less and js files from Bootstrap, not the bundled version. None of the less files exist anymore, and the js files have all moved or are packaged differently.

There are also a significant number of breaking changes between bootstrap 3, 4, and 5, which haven't been addressed.

This is also removing the use of the Bootstrap tooltip component. That isn't tying us to jQuery at all. Bootstrap 5 still includes a tooltip component, which doesn't use jQuery. The things tying us to jQuery are the GitHub popup, table sorting, and the autocomplete dropdown. There are things in the js code using Bootstrap that need to call it via jQuery (like the ttip code being removed in this PR) but those would instead need to be adapted to Bootstrap 5's new calling conventions.

oalders:
I think we could close this for the time being. I don't believe @JordanEdward02 has the bandwidth for it right now.

PR #3339, opened 2025-05-03
Revert exposed -web port from 8000 to 5001

oalders:


haarg (approved):
The reason I had changed the exposed port to be 8000 is that macOS uses port 5000 by default for its own purposes. Technically it is still possible to listen on port 5000 on localhost, but this can get confusing. And obviously port 5001 is not the same and is not occupied, but I was hoping to get away from the 5000/5001 pairing we were using previously. I had missed that GitHub has the port 5001 hard coded in its dev app.

I'm fine with merging this if it makes things easier.

PR #3341, opened 2025-05-03
oalders/stars

oalders:
- **Add GitHub stars to favorite.tx**

This can be tested via the url `/release/OALDERS/HTML-Restrict-v3.0.2`.

Screenshot 2025-05-03 at 1 10 02 PM

Note how badly styled it is. There's no need to keep anything from the current style.

### DIFF
--- a/root/inc/favorite.tx
+++ b/root/inc/favorite.tx
@@ -12,8 +12,8 @@
 </div>
 %% if $github && $repository.url {
   <div class="star-rating">
-    <a href="[% $repository.url %]" target="_blank" class="btn btn-primary">
-      ⭐ Starred <span class="badge badge-primary">[% $github.stars %]</span>
+    <a href="[% $repository.url %]" target="_blank" class="btn btn-primary" title="View on GitHub">
+      ⭐ Starred <span>[% $github.stars %]</span>
       <span class="sr-only">GitHub stars</span>
     </a>
   </div>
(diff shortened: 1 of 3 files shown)


root/inc/favorite.tx
```diff
@@ -10,3 +10,11 @@
 <div class="show-logged-out">
     <button class="fav-not-logged-in favorite[% if $favorites { %] highlight[% } %]"><span>[% $favorites %]</span> ++</button>
 </div>
+%% if $stars {
```
guest20, line 13: Should the control disappear when there are zero stars?

oalders: That was my original thought, but now I'm not sure.

JordanEdward02:
Tried to make it match the favourites button as much as I could, but obviously couldn't go with orange background because the contrast with the star is awful. Please let me know if you think any changes are needed :) 

Screenshot from 2025-05-06 17-54-07
Screenshot from 2025-05-06 17-54-01

guest20:
Perhaps there should be a github icon in the area to indicate that they are github stars?

Also, when talking about github stars the yellow/full star means you, yourself have starred it:
image

This is what the control on github looks like when you haven't starred the thing yourself:
image

oalders:
> Tried to make it match the favourites button as much as I could, but obviously couldn't go with orange background because the contrast with the star is awful. 

Thanks for this! For some reason I missed the notification until now.

> Perhaps there should be a github icon in the area to indicate that they are github stars?

I think there is a font-awesome GitHub icon we could add to this.

> Also, when talking about github stars the yellow/full star means you, yourself have starred it

We don't have a good way of knowing if the user has starred the repo or not, so I don't know that there is a perfect solution here.

PR #3342, opened 2025-05-03
Add an author flag to the sidebar

### LINKED ISSUE #2750: Empty author links in sidebar
For example: https://metacpan.org/author/JGNI

Screen Shot 2022-09-29 at 15 17 09

This author has a GitHub section in their profile with an empty id: https://explorer.metacpan.org/?url=/author/JGNI

Normally the pin would have a location name beside it. In this author's case, there are lon/lat co-ordinates. So, we need to fix how we handle that here.

oalders:
This fixes a display issue when the author has a country set, but no
city or region. Closes #2750

### DIFF
--- a/root/author.tx
+++ b/root/author.tx
@@ -19,6 +19,7 @@
     <li class="author-item">
         <i class="fas fa-map-marker-alt"></i>
     %%  }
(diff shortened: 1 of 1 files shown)


oalders:
Screenshot 2025-09-28 at 9 31 57 PM

PR #3343, opened 2025-05-03
Don't rewrite a search prefix when attempting a suggestion

### LINKED ISSUE #1265: Module name suggestion should preserve original search
I tried using a `dist:` search that wasn't an exact match, and the suggestion uses the translated field instead of the original search term, which seems much less likely to be helpful:

https://metacpan.org/search?q=dist%3AMouseX

> Did you mean : file.distribution::MouseX

We should base the suggestion off of the original query rather than after we've munged the field names.

oalders:
Fixes #1265

- **Add some aria roles to search form and search alert**
- **Fix suggestion handling when there's a search prefix**

### DIFF
--- a/e2e/home.spec.ts
+++ b/e2e/home.spec.ts
@@ -1,6 +1,48 @@
-import { test, expect } from '@playwright/test';
+import { test, expect } from "@playwright/test";
 
-test('home page', async ({ page }) => {
-  await page.goto('/');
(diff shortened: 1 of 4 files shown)


oalders:
@pjcj does this look reasonable?

pjcj:
Looks reasonable to me.  Does it produce reasonable results?

oalders:
@pjcj yes, it does. There is a line of code that I added which only shows up as covered after the Playwright tests have run.

PR #3361, opened 2025-05-24
Change configuration file name

ehuelsmann:
The old instructions are not valid YAML.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,11 +102,11 @@ with `precious path/to/file`
 
 The back end defaults to `api.metacpan.org/v1`. Running a local API server is
 optional and not required to hack on the front end. The address to the API being
-used can be changed in the `metacpan_web.conf` file. Ideally you would create a
-new file called `metacpan_web_local.conf` that contains
+used can be changed in the `metacpan_web.yaml` file. Ideally you would create a
+new file called `metacpan_web_local.yaml` that contains
 
 ```bash
-api                 http://127.0.0.1:5000
+api: http://127.0.0.1:5000
 ```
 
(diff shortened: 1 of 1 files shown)


haarg:
The existing instructions aren't really wrong. Using a `metacpan_web_local.conf` file using `Config::General` syntax should still work.

The file name also needs to be updated with the format.

ehuelsmann:
> The file name also needs to be updated with the format.

The confusion came from the fact that - although the docs list `metacpan_web.conf`, there actually is only a `metacpan_web.yaml`. All I did was add the `_local` suffix to the filename on disk (that is, before the extension), which resulted in `metacpan_web_local.yaml` instead of the documented `metacpan_web_local.conf`.

haarg:
`metacpan_web_local.conf` still needs to be changed to `metacpan_web_local.yaml`

haarg:
I think we should change the example to also use YAML, like you originally had. But both file names need their extensions changed.

I just wanted to clarify that a `_local.conf` will still work. But we don't need it in the docs.

ehuelsmann:
Ah! No problem. Reverted the last change (reinstating `metacpan_web_local.yaml`) and changed the base name of the standard configuration to `metacpan_web.yaml` as it is on disk.

oalders:
> I think we should change the example to also use YAML

+1 Let's not encourage any config formats other than YAML. We were in a transitional state there for a while, but now it's time for the docs to catch up.

haarg (approved):

oalders:
Thanks @ehuelsmann!

PR #3363, opened 2025-05-25
Tell the bots to go away

ranguard:
The site is having issues at the moment, if it's not there for humans then the bots shouldn't be taking up resources.

We can decide if someone these should be allowed in future


oalders (approved):
We can chat more about this at some point in the future. This is a good start for today.

PR #3373, opened 2025-06-02
Docker and GitHub Action updates

haarg:
- smaller image for asset building
  - use timestamp tags for metacpan-base rather than latest
  - add dependabot updates for docker
  - support slim builds (about 1/3 the size)
  - various other small changes


.github/workflows/test.yml
```diff
@@ -77,7 +77,7 @@ jobs:
           load: true
       - name: Run Perl tests
         run: >
-          docker run -d -i -p 8000:80
+          docker run -d -i -p 8000:8000
```
oalders, line 80: I think these may need to be changed to `5001`? I just merged that change to the port number. 😢

oalders (approved):
Just the issue with s/8000/5001/ that may need to be addressed.

Thank you for the descriptive commit messages. 👍🏻

PR #3374, opened 2025-06-04
Better caching and serve from cache more

ranguard:
IF reach the main /end  we enable 

    $c->cdn_stale_while_revalidate('1d');
    $c->cdn_stale_if_error('1y');

These only take effect if `cdn_max_age` has been set

Also added caching on /search - which is where most of our traffic ends up

https://www.fastly.com/documentation/guides/concepts/edge-state/cache/stale/ shows the timeline


oalders (approved):

lib/MetaCPAN/Web/Controller/Pod.pm
```diff
@@ -40,6 +40,7 @@ sub view : Private {
     my $release    = $c->stash->{release};
 
     if ( $data->{directory} ) {
+        $c->cdn_max_age('1y');
```
haarg, line 43: This can end up being a redirect from a non-versioned path to a versioned path, and it doesn't have any surrogate keys so it won't get purged.

oalders (approved):

PR #3380, opened 2025-06-14
Remove favourite function - causing traffic

ranguard:
Getting favourite data on the API is called... A LOT! - this is in 24 hours

Screenshot 2025-06-14 at 13 16 33

This removes fetching/displaying fav information  - I've left it in https://metacpan.org/account/favorite/list in case anyone wants to extract their old data


oalders:
I think this is fine as a temporary measure. I'd like to defer to @haarg on the diff, though.

PR #3387, opened 2025-06-21
Leo/less pagination

ranguard:
So that bots/crawlers do not try large resultsets


guest20:
Don't higher per-page results reduce the per-overhead?

ranguard:
It is more about the number of page variations that bots end up crawling - I didn't change the defaults but might be worth doing that as well... and going from 20-50 or something - not sure

PR #3390, opened 2025-06-25
Add Codeberg to profile options

1-1sam:
[Codeberg](https://codeberg.org/) is a fairly popular alternative code-hosting platform. I think it would be nice if it were to be included with the default profile options.


oalders:
Thanks @1-1sam! Could you include a screenshot of the change?

1-1sam:
I'm having issues getting a screenshot, I can't associate an identity with a local metacpan-web instance. When I try to use a Google account, I get a `redirect_uri_mismatch` error. If I try to use a GitHub account, I just get a `https://fastapi.metacpan.org/user` JSON object. I also had the same issues when running a local metacpan-api instance. Any idea on what might be the issue?

oalders:
There is something funky with the GitHub auth when testing locally. I'm going to merge this and we can test it in production. 😄 Thanks, @1-1sam!

PR #3391, opened 2025-06-26
snapshot update improvements

haarg:
Use signed commits (created via the API) rather than needing to calculate the git user info for the bot.

Fail if fetching cpm doesn't work.

Use perl-actions/get-prereqs rather than a weird cpanm invocation to calculate the forced prereqs.

Install ExtUtils::MakeMaker as part of the forced dependencies to ensure consistency.


ranguard (approved):
looks ok

PR #3460, opened 2025-10-26
use eslint with stylistic for tidying javascript

haarg:
eslint and stylistic is more configurable than prettier and can have better alignment. We are also already using eslint.


haarg:
prettier is still being used to tidy a number of other file types.

PR #3487, opened 2025-12-28
Refactor markdown rendering to be viewed using /view/ URLs

haarg:
This extends the `/view/` controller to recognize markdown files and use a dedicated action and template to view the rendered files. The `/source/` URLs are restored to properly showing the source the markdown files.


oalders:
Screenshot 2026-02-09 at 3 22 51 PM

PR #3524, opened 2026-02-18
Re-organize about section

oalders:
This tries to consolidate some of the "About" information while trying
to make important information more discoverable.


haarg (approved):

oalders:
That test failure was unrelated (RSS feed).

PR #3533, opened 2026-02-24
Try to fix codecov upload permissions for forks

oalders:
- **Don't try to upload to codecov if the token is not available**
- **Try to upload codedov reports even if forked branches**

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,32 @@
+---
+name: Upload coverage to Codecov
+
+on:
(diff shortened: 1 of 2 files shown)


haarg:
This is essentially the same this I was trying to do in #3084 but didn't get working.

oalders:
The other option would be to allow coverage uploads from forks without requiring a key. Or we forego coverage reports on forks. Given how few PRs we get from forks I guess this is probably not worth spending a lot of time on.

PR #3537, opened 2026-02-27
Add a new opengraph image

oalders:
- use a png with 1200x630px dimensions rather than a square
- avoid transparency as it can render in weird ways depending on where
  the image is shared


oalders:
The failure is some weird codecov thing. Doesn't apply to this PR.

PR #3543, opened 2026-02-28
don't detach from sidebar on wide screens

### LINKED ISSUE #2869: Layout is a little funny on wide screens.
This is a VERY edge case. I have a curved monitor and noticed that the left bar kinda detaches from the content when the screen gets wide enough.

Screen Shot 2023-02-20 at 11 54 23 AM

The most distracting part is how the breadcrumbs are left justified.

I notice that Wikipedia handles this correctly. Someone suggested I report this.

Screen Shot 2023-02-20 at 11 55 34 AM

oalders:
Fixes #2869

### DIFF
--- a/root/static/less/breadcrumbs.less
+++ b/root/static/less/breadcrumbs.less
@@ -5,7 +5,7 @@
     display: flex;
     flex-wrap: wrap;
     justify-self: center;
-    max-width: 1200px;
+    max-width: @content-width;
     padding: 20px 20px 0 20px;
     width: 100%;
 
--- a/root/static/less/global.less
+++ b/root/static/less/global.less
@@ -396,7 +396,8 @@ h3,
     grid-template-columns: fit-content(@sidebar-width) 1fr;
     grid-template-rows: max-content max-content 1fr max-content;
     line-height: 1.54em;
-    margin-top: 57px;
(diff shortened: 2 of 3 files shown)


oalders:
Screenshot 2026-02-28 at 2 53 33 PM


Screenshot 2026-02-28 at 2 52 22 PM

haarg:
I'm not clear on what the screenshots are meant to show. They are showing two different behaviors but neither matches the current behavior.

oalders:
That top one shows the "before" on my 2560 x 1440 monitor. Do you see something different?

haarg:
I see what is shown in #2869, where the breadcrumb is left aligned but the rest is centered.

The fix here is probably fine. I would almost prefer what is shown in your "before" screenshot if it was consistent. But I don't use a wide screen so my opinion doesn't really matter.

oalders:
I don't feel too strongly about this. @toddr are you ok with the production layout for these pages? I also don't really have a wide layout. On a big screen I'll have to browser windows side by side, so I'd never even see this most days.

toddr:
That certainly looks better. As I said in the original issue it’s a weird case that even I myself don’t usually have the window that wide. I would say this resolves it.

PR #3617, opened 2026-04-09
DARK MODE

oalders:


Grinnz:
MINOR: Contrast of the favorites and github star buttons are a little jarring
image

Grinnz:
pod2html textarea text is dark gray
image

Grinnz:
Github info popups aren't completely darkified
image
image
image

t/html.t
```diff
@@ -8,7 +8,7 @@ use File::Find ();
 use Path::Tiny qw( path );
 
 # files that have inline <script> tags
-my %skip = map { $_ => 1 } ();
+my %skip = map { $_ => 1 } ('root/base.tx');
```
haarg, line 11: This isn't needed when using an external js file.

root/base.tx
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en-US">
     <head>
-        <script>(function(){var t=localStorage.getItem('theme');if(!t)t=window.matchMedia('(prefers-color-scheme:dark)').matches?'dark':'light';document.documentElement.setAttribute('data-bs-theme',t)})()</script>
+        <script src="/static/js/theme-init.js"></script>
```
haarg, line 4: This could present caching problems if the script needed to be changed. It's simple enough that we may think that's acceptable.

Probably needs `blocking="render"`.

root/static/js/theme-init.js
```diff
@@ -0,0 +1,7 @@
+(function () {
+    var t = localStorage.getItem('theme');
```
haarg, line 2: If this was a `let` rather than a `var`, the wrapping anon function would not be needed.

The file probably wants a `'use strict';`

t/html.t
oalders: I didn't think so either, but the test failed until I added it. I'll test again.

haarg: Looks like the checks the test is doing need to be fixed.

oalders:
I think all of the comments have been addressed. I've also done an a11y audit for contrast which picked up a few things that are now fixed as well.

ranguard (approved):

PR #3627, opened 2026-04-24
Add GFM support

oalders:


lib/MetaCPAN/Web/RenderUtil.pm
```diff
@@ -39,7 +34,7 @@ sub filter_html {
     my $hr = HTML::Restrict->new(
         uri_schemes => [ undef, 'http', 'https', 'mailto', 'irc', 'ircs' ],
         rules       => {
-            a       => [qw( href id target )],
+            a       => [qw( href id target class aria-hidden )],
```
haarg, line 37: Pretty sure we don't want to allow these in user content.

PR #3644, opened 2026-04-30
update dependabot config

oalders:
add cooldown and group updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,20 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    groups:
+      major-updates:
+        patterns:
+          - '*'
+        update-types:
+          - 'major'
+      minor-and-patch:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -15,7 +29,18 @@ updates:
...
       interval: daily
+    groups:
+      minor-and-patch:
+        patterns:
+          - '*'
+        update-types:
+          - 'minor'
+          - 'patch'
+    cooldown:
+      default-days: 7
```
haarg: The only updates for docker that we would be taking is for our own base image, where the cooldown wouldn't be appropriate. I never got it to work successfully though.

haarg (dismissed):

.github/dependabot.yml
oalders: Thanks -- dropped the cooldown prior to merge.

PR #3680, opened 2026-06-24
Add to FAQ: Can I use my PAUSE id to login to MetaCPAN?

szabgab:
Since I posted about it in the Perl Weekly I got several questions about MetaCPAN. It might be a good idea to include the answers explicitly in the FAQ.


oalders:
Thanks, @gabor. With @cpan.org forwarding having been turned off, we need to find a new way to link PAUSE accounts. We have some ideas, but need some time to implement it.

PR #3730, opened 2026-09-17
Tidy imports: enable perlimports sort + blank-line-delimited blocks

oalders:
Enable `sort` in `perlimports.toml` and reorganize `use` statements into blank-line-delimited blocks:

- `use Moo`/`use Moose`/`*::Role` hoisted into their own block above the sorted includes.
- `namespace::clean`/`autoclean` kept in a trailing block.

Blank lines are sort boundaries for perlimports, so the object-system pragma stays put instead of being scattered mid-list. Changes are import reordering/whitespace only — no imports added or removed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/MetaCPAN/Web/Controller/Changes.pm
```diff
@@ -1,7 +1,9 @@
 package MetaCPAN::Web::Controller::Changes;
 
-use Moose;
 use experimental 'postderef';
+
+use Moose;
+
```
haarg, line 4: Moving `use Moose` after `use experimental` will re-enable the experimental warnings.

oalders: Ah, thanks!

ranguard (approved):

